PR #1, opened 2014-11-24
Enable HSTS for all apache served SSL sites

tomhughes:
This enables [HTTP Strict Transport Security](http://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) for all our apache served SSL sites.

This basically means that we tell the browser to default to using https for those sites even if the user types http, until a given expiry date. Currently I have set that 180 days but that is open for discussion.

### DIFF
--- a/cookbooks/apache/recipes/ssl.rb
+++ b/cookbooks/apache/recipes/ssl.rb
@@ -29,6 +29,7 @@ apache_module "socache_shmcb" do
 end
 
 apache_module "ssl"
+apache_module "headers"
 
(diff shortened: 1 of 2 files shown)


KommX:
When will this go live?

tomhughes:
We had to revert it because it breaks too many things. Specially if an OAuth client uses an http URL and the browser silently rewrites it to https then OAuth fails because the request signature is wrong.

So we could only deploy this if every application and web site that uses OAuth with osm.org was using https URLs.

PR #20, opened 2015-04-08
Archive cgimap logs to ironbelly.

zerebubuth:
Currently these aren't archived, meaning we can't investigate issues which occurred more than 7 days ago. Nor can we do a long-term historical analysis of the logs to figure out the best values for, e.g: rate limiting parameters.

### DIFF
--- a/cookbooks/web/templates/default/logrotate.web.erb
+++ b/cookbooks/web/templates/default/logrotate.web.erb
@@ -18,6 +18,7 @@
 <% end -%>
 <% if File.directory?("#{node[:web][:base_directory]}/cgimap") -%>
     /usr/bin/service cgimap reload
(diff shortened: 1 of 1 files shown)


tomhughes:
I didn't mean to archive the damned things forever!

zerebubuth:
What would be an appropriate amount of time? 6 months? 3 months?

tomhughes:
To be clear my thought was that we should maybe increase the expiry to 28 days as for the apache logs, but I haven't looked to see if the disk space on the backend machines is enough for that.

I had forgotten that we also send the apache logs to ironbelly...

As ironbelly is filling up fast I'm reluctant to push too much more too it.

tomhughes:
Compressed cgimap logs are only about 20Mb per day on each machine so extending expiry to 28 days on the live servers shouldn't be an issue.

PR #21, opened 2015-04-24
Slew after 2015 leap second rather than have ntpd flag kernel of change.

Firefishy:


Firefishy:
I need to check if we are effected by: http://bugs.ntp.org/show_bug.cgi?id=2745

tomhughes:
It's one second for crying out loud, is anything really going to care?

Firefishy:
The issue is not the second, it is potential app issues a repeated second could trigger.

tomhughes:
I realise that, but do we have any reason to believe any of our apps will have a problem?

I mean for the most parts things will just deal with unix timestamps (seconds since 1970) and it will just be a second like any other. Only if they convert that to the component parts will they notice that second 60 exists in that minute and then they would have do something (other than print it to a log or whatever) that caused a problem.

Firefishy:
I don't have time to check, but slewing seems like a safer bet to me.
For the 2012 leap second Google’s Site Reliability team decided to slew rather than have the weird kernel step: http://googleblog.blogspot.co.uk/2011/09/time-technology-and-leaping-seconds.html

Firefishy:
There are many reports of apps going 100% CPU usage after the 2012 leap second, traced back to futex_wait.
eg: http://www.redmine.org/boards/2/topics/31731

gravitystorm:
I think this is obsolete now, right?

Firefishy:
Yip, 2015 leap second has come and gone. Closing.

PR #31, opened 2015-08-05
GPX import for dev APIs

zerebubuth:
This attempts to make a definition (LWRP?) for the GPX import daemon, which would allow it to be used on each of the development APIs as well as the main API servers, making testing against those APIs easier.

### DIFF
--- a/cookbooks/dev/recipes/default.rb
+++ b/cookbooks/dev/recipes/default.rb
@@ -137,6 +137,8 @@ if node[:postgresql][:clusters][:"9.1/main"]
     database_name = details[:database] || "apis_#{name}"
     site_name = "#{name}.apis.dev.openstreetmap.org"
     rails_directory = "/srv/#{name}.apis.dev.openstreetmap.org"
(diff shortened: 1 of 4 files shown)


Firefishy:
Still pending?

zerebubuth:
Uhh... I don't remember what state this was in. Feel free to replace / rewrite / reject it.

tomhughes:
Well there's a lot of stuff here, and some of it will need work (we probably don't want any new definitions) but the biggest issue is I need to think about the best place to actually put the traces.

That's really the reason I've never been bothered, because I need to work out where to store them and whether we want to have any expiry or anything.

mmd-osm:
This PR can be closed, GPX import daemon got replaced in the meantime and is no longer needed. See https://master.apis.dev.openstreetmap.org/user/TomH/traces/919

tomhughes:
Yes GPX import is now enabled using the new rails job queue based implementation on master.apis.dev.openstreetmap.org and any other branch which has that code.

westnordost:
👍

PR #34, opened 2015-08-24
Allow pinning kernel versions to work around hardware / driver issues.

zerebubuth:
Sometimes hardware issues or driver issues are fixed or broken in specific kernel versions, so it's handy to be able to pin the version to a known-working one while the problem is sorted out or a proper fix is found.

This does this by ensuring that the particular kernel version package is installed and that the grub configuration boots that as the default.

For context, see:

https://github.com/openstreetmap/operations/issues/45


tomhughes:
Merged as d1be18e9426db54d3cf367855ee95fd8ed117e3b and improved a bit in 1df3edbee59a60ca0bd7005124c12e99205fd2d1.

PR #37, opened 2015-10-01
Deconfigure smaug to make way for a new machine.

zerebubuth:
Smaug was moved back to UCL and is currently not powered up. Meanwhile, smaug's IP has been re-used for another machine which I can't log into because smaug's SSH key is still in known_hosts.


tomhughes:
I deleted the smaug node from chef which should cause it to drop out of known_hosts.

PR #43, opened 2015-11-23
Added sanity checks to changeset replication

zerebubuth:
Trying to track down why sometimes zero-size or invalid files are produced. (See zerebubuth/openstreetmap-changeset-replication#2 for background)


tomhughes:
Oh that's easy - was this last Thursday morning? When I rebooted ironbelly ;-)

tomhughes:
Oh no, that's ages ago.

PR #45, opened 2015-12-31
Add unattended upgrades support

Firefishy:
Bring unattended-upgrades support into chef. Many of the servers already have unattended-upgrades installed and partially configured.

By default enable, specifically disable on DB role machines.

Also added "Unattended-Upgrade::Remove-Unused-Dependencies" support, which is primarily used for removing unused kernel packages. (automatically keeps current running kernel + last run + latest)

### DIFF
--- a/cookbooks/apt/attributes/default.rb
+++ b/cookbooks/apt/attributes/default.rb
@@ -1,1 +1,3 @@
 default_unless[:apt][:sources] = []
+default[:apt][:unattended_upgrades][:enable] = true
+default[:apt][:unattended_upgrades][:remove_unused_dependencies] = false
--- a/cookbooks/apt/recipes/default.rb
+++ b/cookbooks/apt/recipes/default.rb
@@ -19,6 +19,7 @@
 
 package "apt"
 package "update-notifier-common"
+package "debconf-utils"
 
 file "/etc/motd.tail" do
   action :delete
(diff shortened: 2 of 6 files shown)


tomhughes:
I never bothered with this because I thought Ubuntu enabled it by default (for security updates only) and I'd rather not do it for other updates.

tomhughes:
Does it really need a non-idempotent step and a seed response file? Can we not just let it install with the defaults and then overwrite the configuration like for every other package?

Firefishy:
This only enables security updates. Quick inspection the following machines do not have unattended-updates installed or enabled:
* drogon.openstreetmap.org
* dulcy.openstreetmap.org
* grindtooth.openstreetmap.org
* kessie.openstreetmap.org
* konqi.openstreetmap.org
* poldi.openstreetmap.org
* sarkany.openstreetmap.org
* yevaud.openstreetmap.org

Firefishy:
The reason I use the non-idempotent apt reconfig + response_file step instead of just writing out the template is because the configure step of the unattended-upgrades deb file manages the /etc/apt/apt.conf.d/20auto-upgrades file by copying over /usr/share/unattended-upgrades/20auto-upgrades or /usr/share/unattended-upgrades/20auto-upgrades-disabled depending debconf "unattended-upgrades unattended-upgrades/enable_auto_updates" state. The apt reconfig sets the correct debconf state and is future proof.

tomhughes:
So basically this would work just as well:

```
package "unattended-upgrades"

file "/etc/apt/apt.conf.d/20auto-upgrades" do
  user "root"
  group "root"
  mode 0644
  contents IO.read("/usr/share/unattended-upgrades/20auto-upgrades")
end
```

and would, in my mind at least, be a lot simpler...

Firefishy:
Yes, that is simpler, but does not set the debconf state. Without the debconf state it isn't potentially future proof against package upgrades or changes.

tomhughes:
So chef will just overwrite it again the next time it runs ;-)

Firefishy:
How do I get contents IO.read to read the correct file depending on the [:apt][:unattended_upgrades][:enable] state?

tomhughes:
Don't worry - I just pushed 0a20dc708587ac216694ef9b3e0632035038cb6e to do it.

If we decide we need to set the debconf setting the best solution will be a `debconf` LWRP that can idemtpotently set a debconf variable.

PR #50, opened 2016-01-17
Add imagery

Firefishy:


Firefishy:
Still WIP. Will finish off over next few days.

PR #51, opened 2016-01-18
Replace characers which aren't allowed in XML 1.0.

zerebubuth:
Characters in the range `\x01-\x20` except for `\t\n\r` aren't allowed in XML 1.0, so we do the same thing as the planet dump code and replace them with '?'.

### DIFF
--- a/cookbooks/planet/files/default/replication-bin/replicate-changesets
+++ b/cookbooks/planet/files/default/replication-bin/replicate-changesets
@@ -18,7 +18,7 @@ GEO_SCALE = 10000000
 ##
 # replace characters which cannot be represented in XML 1.0.
 def xml_sanitize(str)
-  str.gsub(/[\x00-\x08\x0b\x0c\x0e-\x20]/,'?')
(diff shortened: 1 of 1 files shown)


Firefishy:
Really no way to encode these characters in XML?

zerebubuth:
Not in [XML 1.0](https://www.w3.org/TR/REC-xml/#charsets). As @pnorman [pointed out](https://github.com/zerebubuth/openstreetmap-changeset-replication/issues/3#issuecomment-171850693), they're allowed in XML 1.1 (apart from nul), but XML 1.1 isn't widely supported.

zerebubuth:
Although, even in XML 1.1, the using those characters is ["discouraged"](https://www.w3.org/TR/xml11/#charsets).

tomhughes:
Isn't the correct solution to stop the API accepting them, and to fix any that already exist in the database? I thought that was what we already did with other things?

zerebubuth:
Sure, part of the fix is stopping the API accepting them. But these things have a way of creeping back into the database, so part of the fix is not producing invalid XML when they do.

PR #56, opened 2016-02-16
Reduce tile log analysis delay.

zerebubuth:
Take yesterday's squid logs for analysis rather than two days ago.

@Firefishy - I think there was a reason why we were taking 2 days ago, but I can't figure it out now. Seems to me that this would work okay, as the logs should be rotated at 0625, compressed and uploaded, then the analysis will run at 1722. Ten hours to compress and upload should be enough, right?

### DIFF
--- a/cookbooks/tilecache/templates/default/logrotate.squid.erb
+++ b/cookbooks/tilecache/templates/default/logrotate.squid.erb
@@ -12,6 +12,6 @@
   sharedscripts
   postrotate
     test ! -e /var/run/squid.pid || /usr/sbin/squid -k rotate
-    /usr/bin/rsync /var/log/squid/zere.log.2.xz ironbelly::logs/tile.openstreetmap.org/<%= node[:hostname] %>-`date -d "-2 days" +%Y-%m-%d`.xz || true
+    /usr/bin/rsync /var/log/squid/zere.log.1.xz ironbelly::logs/tile.openstreetmap.org/<%= node[:hostname] %>-`date -d "-1 days" +%Y-%m-%d`.xz || true
   endscript
 }
--- a/cookbooks/tilelog/templates/default/tilelog.erb
+++ b/cookbooks/tilelog/templates/default/tilelog.erb
@@ -4,7 +4,7 @@ ANALYZE=<%= @analyze_bin %>
 LOGDIR=<%= @input_dir %>
 OUTDIR=<%= @output_dir %>
 if [ -z "$DATE" ]; then
-		DATE=`date -d "2 days ago" "+%Y-%m-%d"`
+		DATE=`date -d "1 day ago" "+%Y-%m-%d"`
 fi
 TMPDIR=`mktemp -d -t tmp.XXXXXXXXX`
 ORIGDIR=`pwd`


tomhughes:
Do they actually get compressed the same day? A lot of our logs are set to rollover one day and compress the next, so 1.xz would never exist...

zerebubuth:
I'm not 100% sure, but my reading of the [logrotate config](https://github.com/zerebubuth/openstreetmap-chef/blob/reduce-squid-log-delay/cookbooks/tilecache/templates/default/logrotate.squid.erb) is that it should compress immediately. I don't see a `delaycompress` in the config, and I think the default is to not delay compression. I checked on one server and there was a `.1.xz`, but I don't know if that was created immediately upon rotation or was backfilled later.

tomhughes:
You may well be right. I know the apache logs on the main site are delayed, and I suspect that the logic was simply copied over from there when we started storing the tile logs.

Firefishy:
+1 Looks good.

tomhughes:
So nepomuk, jakelong, drogon and katie all failed last night:

```
/etc/cron.daily/logrotate:
rsync: link_stat "/var/log/squid/zere.log.1.xz" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1183) [sender=3.1.0]
```

zerebubuth:
As far as I can tell from [the logrotate docs](http://linuxcommand.org/man_pages/logrotate8.html),

> The **sharedscripts** means that the **postrotate** script will only be run once(after the old logs have been  compressed), not  once  for  each  log which is rotated.

The `zere.log.1.xz` file exists on all those hosts now, (although katie seems to have created the file at 05:28, almost an hour before the cron.daily runs?) which makes me think that the file hadn't been compressed when the error was generated... Is there a timestamp on that logrotate/rsync error?

tomhughes:
So in fact they all failed in the end I think, just some of them took longer because the logs are larger.

The timestamp is because it preserves the time of the original log by updating it once the compression finished. I watched some of them while they were running (in fact gorynych is still running now) and it definitely seems to compress before the rsync so I have no idea what is going on...

tomhughes:
I really don't understand this... I ran inotifywait on gorynych today to watch the log directory and this is the log of events:

```
2016-02-23 09:40:14 /var/log/squid/ OPEN zere.log.1
2016-02-23 09:40:14 /var/log/squid/ CREATE zere.log.1.xz
2016-02-23 09:40:14 /var/log/squid/ OPEN zere.log.1.xz
2016-02-23 12:00:29 /var/log/squid/ CLOSE_WRITE,CLOSE zere.log.1.xz
2016-02-23 12:00:30 /var/log/squid/ DELETE zere.log.1
2016-02-23 12:00:30 /var/log/squid/ CLOSE_WRITE,CLOSE zere.log.1
2016-02-23 12:00:30 /var/log/squid/ OPEN zere.log.3.xz
2016-02-23 12:00:30 /var/log/squid/ DELETE zere.log.3.xz
2016-02-23 12:00:30 /var/log/squid/ CLOSE_WRITE,CLOSE zere.log.3.xz
```

Which is pretty much as you would expect, but that the email arrives, timed at 12:00:31 saying that it couldn't find zere.log.1.xz which we've seen was closed two seconds earlier when the compression completed!

zerebubuth:
Is this related? https://bugzilla.redhat.com/show_bug.cgi?id=429873#c17

If so, might it be worth trying the `s/postrotate/lastaction/` suggested fix, or the work-around that the reporter says is working; a `sleep 600` delay? The latter is pretty horrible, so hopefully the former works.

tomhughes:
Moving to lastaction seems to have fixed this.

PR #62, opened 2016-04-25
link new 'Wiki content license' page in the footer

harry-wood:
The current wiki footer link, takes us to http://wiki.openstreetmap.org/wiki/OpenStreetMap_License  which no longer has any information about the wiki content license, so potentially misleading. We've created a new page 'Wiki content license' to be a better destination for that footer link . Some more discussion here: http://wiki.openstreetmap.org/wiki/Talk:OpenStreetMap_License

Swapping this `$wgRightsPage` value should do the trick


tomhughes:
Merged as 05b1a4401d6bff629f92914d312c82a51a750bd3.

PR #65, opened 2016-05-21
Fix users agreed script.

zerebubuth:
Users agreed script was modified to use the setting for the current database, which means it is a template now. So it needs to be run from the recipe.


tomhughes:
Already fixed in ee2ccdb351eeaa09fde5a9667958cac94977fa0c.

PR #69, opened 2016-06-20
Use PPA for cgimap

zerebubuth:
Instead of building `cgimap`, which requires a compiler and a whole bunch of other unnecessary things and takes a few minutes, instead install from a pre-compiled PPA package.


cookbooks/web/recipes/base.rb
```diff
@@ -17,13 +17,6 @@
 # limitations under the License.
 #
 
-node.set[:nfs]["/store/rails"] = {
-  :host => node[:web][:fileserver],
-  :path => "/store/rails"
-}
-
-include_recipe "nfs"
-
```
tomhughes: What does this have to do with using a PPA for cgimap?

cookbooks/web/attributes/default.rb
```diff
@@ -1,3 +1,6 @@
 default[:web][:base_directory] = "/srv/www.openstreetmap.org"
 default[:web][:pid_directory] = "/var/run/web"
 default[:web][:log_directory] = "/var/log/web"
+
+# need the APT source for cgimap to be able to install the package
+default[:apt][:sources] = node[:apt][:sources] | ["openstreetmap-cgimap"]
```
tomhughes: Can we put this in the recipe please so that only machines using cgimap get it? I think there are other examples where we do that...

cookbooks/web/templates/default/cgimap.init.erb
```diff
@@ -0,0 +1,56 @@
...
+if node[:web][:readonly_database_host]
+  database_host = node[:web][:readonly_database_host]
+  database_readonly = true
+else
+  database_host = node[:web][:database_host]
+  database_readonly = node[:web][:status] == "database_readonly"
+end
+
+memcached_servers = node[:web][:memcached_servers] || []
+%>
```
tomhughes: Can we leave all this in the recipe please and just pass the variables in as I don't really like having large chunks of code embedded in templates like this?

cookbooks/web/recipes/base.rb
zerebubuth: I was testing the cookbook locally, and it was difficult given that I don't have an NFS server locally to point it to. So I moved that bit out of `base` into `nfs` and included that from the `rails` and `gpx` recipes where it's actually used.

cookbooks/web/attributes/default.rb
zerebubuth: Okay. I saw this in the `passenger` recipe, and figured that it was the way to do it. I'll look for other examples.

tomhughes:
So "big picture" questions...

What are the advantages of using a separate PPA for this over the existing OSM PPA?

Does the PPA package include an init script? Only I see you now have the whole thing as a template so we no long get upstream changes, plus if it's overwriting one from the package then that is problematic.

cookbooks/web/recipes/base.rb
tomhughes: Separating it out may well be a good idea, but it should be a separate commit really.

cookbooks/web/templates/default/cgimap.init.erb
zerebubuth: Sure. Fixed in 54a8cfe.

cookbooks/web/recipes/base.rb
zerebubuth: Reverted in 3f4c1a0.

cookbooks/web/attributes/default.rb
zerebubuth: Another example, `logstash-forwarder`, includes `[:apt][:sources] = ["logstash-forwarder"]` in its default attributes in `roles/logstash-forwarder.rb`. Is it safe to do that in `web-backend.rb`, or will one or other set of default attributes override the other, rather than adding together?

zerebubuth:
> What are the advantages of using a separate PPA for this over the existing OSM PPA?

I'm able to push packages to the separate PPA, but not the OSM one.

> Does the PPA package include an init script? Only I see you now have the whole thing as a template so we no long get upstream changes, plus if it's overwriting one from the package then that is problematic.

There's an example init script included with the documentation, but the cgimap package doesn't install it to `/etc`. Generating the init script here won't overwrite anything upstream.

A lot of the stuff in the init script seems to be configuration, and perhaps would be better extracted to a config file. I could add support for a config file, and then the init script would be more uniform and we could standardise one from the package? Otherwise, we'll just end up heavily modifying it, which could potentially lead to breakage if there were incompatible upstream changes.

cookbooks/web/attributes/default.rb
tomhughes: Yes you could put it in the role. Seems I was wrong about having ones which did it in the recipe having now looked...

tomhughes:
Well we can always give you access to the existing PPA...

As and when we move those machines to 16.04 we'll likely want to replace it with a systemd unit anyway. A config file certainly wouldn't hurt though.

zerebubuth:
Oh, are the backend machines still on 14.04? I haven't yet made a package for those, so this isn't ready to merge.

Also, I tried out putting the PPA in the role, as in 95b2acd, but I'm still worried that it'll override the role for `logstash-forwarder` and not enable that repo. Perhaps it would be worth changing that from a list of repos to enable into a `Hash` of repo name to boolean flag? That would be composable between roles/cookbooks as long as the names were unique. What do you think?

tomhughes:
Yes I'm waiting on an official build of passenger before upgrading the web servers.

tomhughes:
Where are we at with this? The servers are all upgraded now. Am I right in thinking you were going to build cgimap into the admins repo rather than using a separate one?

zerebubuth:
Yup. I've uploaded cgimap 0.5.0 to the osmadmins repo. There's currently 41 hours worth of jobs in launchpad's queue, so we might have to wait a while.

cookbooks/web/recipes/cgimap.rb
```diff
@@ -124,7 +57,7 @@
   service "cgimap" do
     action [:enable, :start]
     supports :restart => true, :reload => true
-    subscribes :restart, "execute[cgimap-build]"
+    subscribes :restart, "dpkg_package[openstreetmap-cgimap-bin]"
```
tomhughes: Is `dpkg_package` required here? or will `package` work? Not sure we have any previous example...

tomhughes: Actually we do and it uses `dpkg_package` so I guess this is right.

PR #74, opened 2016-07-28
Add test-kitchen

gravitystorm:
This initial configuration uses kitchen-vagrant to test the tools cookbook.

I tried, and failed^Whaven't yet succeeded in getting more complex cookbooks to run, so I fell back on a pretty simple cookbook. Even then, there were challenges getting it to work with kitchen-docker, so I stuck with kitchen-vagrant for now.

The choice of a global test-kitchen setup vs a per-cookbook setup is not my first preference. Unfortunately there's no straightforward way to implement anything else, given our mono-repo and not using any cookbook publication systems. Using a global test-kitchen config is also the result that @zerebubuth came up with independently during work in June.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.kitchen/
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -1,0 +1,14 @@
+---
+driver:
+  name: vagrant
+
+provisioner:
+  name: chef_zero
+
+platforms:
+  - name: ubuntu-16.04
+
+suites:
+  - name: tools
+    run_list:
+      - recipe[tools::default]
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source "https://rubygems.org"
 
 gem "rubocop"
 gem "foodcritic"
+gem "test-kitchen"
+gem "kitchen-vagrant"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
+    artifactory (2.3.3)
     ast (2.3.0)
     cucumber-core (1.5.0)
       gherkin (~> 4.0)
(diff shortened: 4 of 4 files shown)


gravitystorm:
Oh, I should note that the "testing" of the tools cookbook only goes as far as the recipe converging successfully, it doesn't yet have any serverspec-based verification.

tomhughes:
Hmm, so this seems to want to use virtualbox which is something that is not getting anywhere near my machine... Not sure if it is explicitly requesting that somehow or not, as I thought I had vagrant using libvirt...

tomhughes:
Ah you need to set `VAGRANT_DEFAULT_PROVIDER=libvirt` but then you find that the bento/ubuntu-16.04 box doesn't support the libvirt provider!

gravitystorm:
It's possible to override configuration using a .kitchen.local.yml file, if that's useful. 

https://jjasghar.github.io/blog/2014/02/17/dot-kitchen-dot-local-dot-yml-and-when-you-want-to-use-it/

That should allow setting the driver provider, and will hopefully allow you to choose a different image.

tomhughes:
Yes I've managed to get it working now with the peru/ubuntu-16.04-server-amd64 box.

It's pretty slow mind - abut 3 minutes to run that one test on my machine.

tomhughes:
Just for the record this `.kitchen.local.yml` worked for me:

```yaml
---
driver:
  provider: libvirt

platforms:
  - name: ubuntu-16.04
    driver:
      box: peru/ubuntu-16.04-server-amd64
```

mmd-osm:
I really wondered how you test all those cookbooks and stumbled across this old pull request. Running anything non-trivial seems to be quite challenging due to lots and lots of dependencies (hardware, networking, hostnames, ... just to name a few).

> Even then, there were challenges getting it to work with kitchen-docker, so I stuck with kitchen-vagrant for now.

I've changed my .kitchen.yml to run with docker and Ubuntu 18.04 running on a Ubuntu 18.04 host. This seems to be quite ok for the purpose of testing easy recipes, and performance isn't as bad as with virtualbox.

```
---
driver:
  name: docker
  provision: true
  vagrantfiles:
    - .kitchen.provision.rb
  provision_command: mkdir -p /run/sshd

provisioner:
  name: chef_zero
  roles_path: roles
#  log_level: info

platforms:
  - name: ubuntu-18.04
```



systemd causes some serious screw up with Docker on Ubuntu 18.04, which could be resolved by adjusting the following line in /etc/default/grub and reboot (new systemd parameter added):

```
 GRUB_CMDLINE_LINUX_DEFAULT="quiet systemd.legacy_systemd_cgroup_controller=yes"
```

PR #77, opened 2016-08-10
Block tile abusers.

zerebubuth:
These are currently 7% of total usage between them:

1. User-Agent `Mozilla/5.0 (Windows NT 5.1)` is not a real browser.
2. User-Agent `osmdroid` is a library, but the app should set this to an specific identifiable value.
3. When no User-Agent or referer is set.

### DIFF
--- a/cookbooks/tilecache/templates/default/squid.conf.erb
+++ b/cookbooks/tilecache/templates/default/squid.conf.erb
@@ -24,7 +24,7 @@ acl osmtileScrapers browser MSIE.5.5
 acl osmtileScrapers browser ^LoadOSM\.exe$
 acl osmtileScrapers browser ^app_name$
 acl osmtileScrapers browser ^osmdroid$ # app using osmdroid library not setting app-specific User-Agent
-acl osmtileScrapers browser ^Mozilla/5.0 (Windows NT 5.1)$ # Faked User-Agent
+acl osmtileScrapers browser ^Mozilla/5\.0 (Windows NT 5\.1)$ # Faked User-Agent
 
 http_access deny osmtile_sites osmtileScrapers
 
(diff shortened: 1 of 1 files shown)


zerebubuth:
@Firefishy what do you think?

cookbooks/tilecache/templates/default/squid.conf.erb
```diff
@@ -23,6 +23,8 @@ acl osmtileScrapers browser MSIE.7\.0.*Windows.NT.5\.1.*2\.0\.50727.$
 acl osmtileScrapers browser MSIE.5.5
 acl osmtileScrapers browser ^LoadOSM\.exe$
 acl osmtileScrapers browser ^app_name$
+acl osmtileScrapers browser ^osmdroid$ # app using osmdroid library not setting app-specific User-Agent
+acl osmtileScrapers browser ^Mozilla/5.0 (Windows NT 5.1)$ # Faked User-Agent
```
Firefishy: Needs escaping. Specifically the .

Firefishy:
Besides the escaping issue on regex, looks good. :thumbsup:

zerebubuth:
Pushed this to chef just now.

zerebubuth:
I had to comment all this back out again, as it had blocked everything. As a result, we served no tiles for between 30 and 60 minutes, depending on how quickly I was able to re-run `chef-client`. We either need some way to test these rules before going live with them, or we need to re-check them for proper escaping, etc...

gravitystorm:
`acl has_referer referer_regex .`

By my reading, that matches referers with exactly one character, rather than `.+` meaning one or more characters. If so, that means that we blocked all traffic unless it had both a referer with exactly one character and a useragent with exactly one character.

tomhughes:
Well no because it's not anchored so it will match anything with at least one character surely?

gravitystorm:
Fair point

PR #79, opened 2016-08-11
Delay browsers not sending referer headers

zerebubuth:
Add more delay pools of smaller size to use for browsers which don't send referer headers.

Also re-wrote the many, many lines of config to use ERB. I found this to be necessary for understanding... but it's possible it was done the other way for a very good reason. @Firefishy - should I expand out the ERB explicitly?

Follows on from #78 as a way of defending against malicious use of the [no-referer setting](https://www.w3.org/TR/referrer-policy/).

I don't know Squid that well, so I'd really appreciate help and reviews of this. Hopefully we can avoid a repeat of the over-blocking yesterday.


Zverik:
There is a suspicious comment "This doesn't seem to work! Revisit later!" — maybe it should be tested first on a development server.

zerebubuth:
That was the old config line which I thought was responsible for blocking everything yesterday. As `daniel_hozac` mentioned on IRC yesterday, it's more likely to have been the unescaped space in [a previous rule](https://github.com/zerebubuth/openstreetmap-chef/blob/17fea9276c93b655d30063d70371e4e5a7a00619/cookbooks/tilecache/templates/default/squid.conf.erb#L27) which is currently also commented out.

Testing would be great! Could you help with that, please?

Zverik:
Alas I can't: I don't have Squid on any of my servers.

The theory about unescaped spaces seems plausible: a similar line above that has spaces replaced with `.*`.

Firefishy:
Not near a real connection to test properly, but looks reasonable to me.

gravitystorm:
I've been working on getting the cookbooks running with test-kitchen, so that everyone is able to test them locally using virtual machines. But it's still a work in progress (see #84 among others)

pnorman:
Is it worth me sending out a reminder to dev@ that applications/sites need to either be sending a user-agent identifying the application or a referer?

zerebubuth:
@pnorman, yes, please! Ideally, non-browser apps would send _both_. There's a lot of entries in the log identifying "OSMDroid", "libwww-perl", "Dalvik", "Java" which all seem to be default User-Agent strings, and together (without referer) account for 11% of daily usage.

pnorman:
> @pnorman, yes, please!

[Sent](https://lists.openstreetmap.org/pipermail/dev/2016-August/029409.html). I don't think we need to wait for any responses because we're not changing the usage policy, just automatically rate limiting those who don't follow it.

zerebubuth:
@pnorman that's great, thanks! Agreed about the timing - this is enforcement of the policy rather than a change to it. I want to try and test this config to make sure that it only blocks the things it's intended to. If anyone could help with that, it would be greatly appreciated.

zerebubuth:
I tried testing it with a hacked around version of a VM from https://github.com/gravitystorm/chef/pull/1, but I'm not getting the responses I expected. When trying with a referer, I was able to do 100 reqs in 152s, but without 100 reqs in 277s. That's 82% slower, but I was expecting 10x slower, given how the config is set up.

I was doing this on a VM with `tile.openstreetmap.org` as the upstream parent, so there _are_ networking delays. And Squid delay pools only come into effect for misses, so I requested random tiles in an attempt to make sure most were misses from my local cache.

Does that sound like it's working? I'm still unsure.

Zverik:
Is it live? Can we test it on the production? :)

zerebubuth:
I did a longer run, 650 requests, and the gap is closed - only 19% slower using no referer. I don't think it's doing what I wanted it to :disappointed: 

@Zverik no, not live yet. I'm not confident enough that it's doing the right thing.

zerebubuth:
I refined my test VM, mainly by commenting out [the line which allows localhost unlimited access](https://github.com/zerebubuth/openstreetmap-chef/blob/ce1d25a30db69e155defd80ada0df50a084f164f/cookbooks/tilecache/templates/default/squid.conf.erb#L84), and I'm getting the results expected: With referer, 21s & without 211s.

Firefishy:
The localhost line is required in prod, so that nginx layer doesn't get rate limited.

zerebubuth:
Oh, yeah, sorry - I wasn't suggesting it become part of this patch. But I thought I'd record it for posterity if anyone else wanted to set up a test VM to verify that delay pools are doing what's expected.

tomhughes:
Do we think this is ready to merge then?

tomhughes:
Do we need to update https://github.com/openstreetmap/chef/blob/master/cookbooks/munin/files/default/plugins/squid_delay_pools at all for this? Do we want separate graphs (if possible) for the two sets of delay pools?

zerebubuth:
The changes in f84e757 split the graphs into two; one for the "normal" delay pools, which is mostly requests with a referer, and the other for the "no referer" delay pools.

I think this is mergeable, as long as the last change looks okay?

pnorman:
Munin shows the delay is working and avg and max dropped tiles are about 30% of what they were, tiles are fresher, and overall the metrics show an improved experience for mappers.

PR #82, opened 2016-08-11
Remove networking recipe from munin::default

gravitystorm:
This removes the networking::default recipe from munin::default. It appears to be unnecessary but I'm happy to be told otherwise! The cookbook dependency remains so that munin::default can use the firewall_rule providers.

This also adds test-kitchen configuration for munin::default, along with a serverspec based test that ensures the munin-node service is up and running after the chef-run converges.


tomhughes:
Merged as eb1313a389c4d7f517b78a0dfbe0f81f2d9095a2 and 7e87085093c7714b71216c0c523e2f2e8676c324.

tomhughes:
I had to revert this as it broke munin completely - without that recipe included the clients don't manage to configure the firewall rule that allows the server to access them.

tomhughes:
So the problem here is the slightly odd way that the `firewall_rule` definition works, which relies on the shorewall rules resource having already been declared but not actually having been run yet and patching the variables values on the resource.

I think there is a way to do this better, but I'm still thinking about it in between cursing shorewall for not doing dropin directories.

tomhughes:
So I reworked `firewall_rule` in 2714460f8457266fec0f5f9d72d900b384d5b724 and was then able to reapply this as 6922bbdb6fbb5c8af223c1117d6d7e0b2d41f0ad.

gravitystorm:
Thanks!

PR #89, opened 2016-09-07
More explicit syncing, and sync the directories too.

zerebubuth:
More explicit syncing, and sync the directories too. This attempts to avoid more data loss and empty files. Also adds an individual state file for each replication file.

This might be overkill, but I'm getting pretty tired of this problem coming back every time ironbelly crashes.

### DIFF
--- a/cookbooks/planet/files/default/replication-bin/replicate-changesets
+++ b/cookbooks/planet/files/default/replication-bin/replicate-changesets
@@ -118,6 +118,27 @@ class ChangesetBuilder
 end
 
 ##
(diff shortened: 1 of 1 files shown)


tomhughes:
That is actually a backwards step as it doesn't force the file to disk until after it has been renamed, so there is a window where it exists with the correct name but may get wiped on a reboot.

The critical thing is to make sure the data is really on disk before it is published until the real name...

zerebubuth:
Good point - added even more syncing in bd63b8d.

PR #91, opened 2016-09-15
Swap wiki sign up to use google ReCaptchaNoCaptcha

harry-wood:
Swap to the more modern google No CAPTCHA reCAPTCHA away from the older google reCAPTCHA which gathers address data for google maps.

Fixes: https://github.com/openstreetmap/operations/issues/19

This simple config swap assumes...
* The recent ConfirmEdit version we have installed (1.4) comes with all the necessary bits to enable this captcha class
* Google's old ReCaptcha uses the same API keys as it's new one (anyone know if this is true?)

risky assumptions? well we could make the swap then swap it back again if wiki sign ups are broken. ...or test in a better way somehow

TODO: in future we might figure out a solution which _also_ doesn't request any javascript from google, but really... not contributing to google maps with every OSM wiki sign up would be a step in the right direction

### DIFF
--- a/cookbooks/mediawiki/templates/default/mw-ext-ConfirmEdit.inc.php.erb
+++ b/cookbooks/mediawiki/templates/default/mw-ext-ConfirmEdit.inc.php.erb
@@ -1,7 +1,8 @@
 <?php
 # DO NOT EDIT - This file is being maintained by Chef
-wfLoadExtensions( array( 'ConfirmEdit', 'ConfirmEdit/ReCaptcha' ) );
-$wgCaptchaClass = 'ReCaptcha';
+wfLoadExtensions( array( 'ConfirmEdit', 'ConfirmEdit/ReCaptchaNoCaptcha' ) );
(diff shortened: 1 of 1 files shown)


harry-wood:
BTW I was just reminded of this because I noticed some folks reporting [some other weird problem with the old reCaptcha](https://help.openstreetmap.org/questions/50567/wiki-accountpage-creation-error?page=1&focusedAnswerId=50704#50704). Maybe this'll fix that issue too.

tomhughes:
Does `$wgReCaptchaSendRemoteIP = true` mean we are sending the user's IP on to google?

Other than that I guess this is mostly one for @Firefishy to review...

harry-wood:
It would mean that yes. I tend to think that's the sensible way to set it, as discussed here: https://github.com/openstreetmap/operations/issues/19#issuecomment-156414181 but ...meh

Firefishy:
Merged as 7826864b0655cd92c629a42d7a8dc8914eeb485e

PR #95, opened 2016-10-04
Forward authentication headers on to the backends

zerebubuth:
So that they can check them. The default behaviour appears to be that [it strips out auth-related headers](https://httpd.apache.org/docs/current/mod/mod_proxy_http.html#env), which would be appropriate if it were forwarding information over the internet, but it isn't.

I suspect this may be why cgimap is having trouble detecting the authentication headers.

### DIFF
--- a/cookbooks/web/templates/default/apache.frontend.erb
+++ b/cookbooks/web/templates/default/apache.frontend.erb
@@ -176,6 +176,11 @@
   ProxyTimeout 3600
 
   #
+  # Forward authentication headers to the backends
+  #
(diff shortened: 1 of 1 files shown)


tomhughes:
That (per http://httpd.apache.org/docs/2.4/mod/mod_proxy_http.html) forwards proxy authentication credentials, which are normally used on forward (outbound) proxies not reverse proxies. Certainly there is no proxy authentication involved in our setup.

tomhughes:
Confirmed from the source that this only affects the `Proxy-Authorization` header not the main `Authorization` header.

PR #100, opened 2016-10-27
Output TagInfo <script> tag

harry-wood:
For the "[taginfo taglists](https://wiki.openstreetmap.org/wiki/Taginfo/Taglists)" on the wiki. This thing is already in place and working, but so far we've kept a chunk of code on https://wiki.openstreetmap.org/wiki/MediaWiki:Common.js and @joto has been asking wiki admins (like me) to make little tweaks to it. He wanted to host it at his end instead (good idea?)

The approach for having mediawiki spit out a &lt;script> tag is [based on this answer](http://stackoverflow.com/a/26436873/338265). I have tested it. It works. But maybe you can see a better way.

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -327,7 +327,7 @@ $wgReadOnly = "<%= @mediawiki[:site_readonly] %>";
 $wgHooks['BeforePageDisplay'][] ='outputTaginfoScriptTag';
 function outputTaginfoScriptTag( OutputPage &$out, Skin &$skin )
 {
-    $script = '<script type="text/javascript" src="http://taginfo.openstreetmap.org/js/taglists.js"></script>';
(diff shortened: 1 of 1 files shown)


joto:
One thing: Should the link go to https? Or `//taginfo.openstreetmap.org` so it works with both http and https?

harry-wood:
oh yeah. well spotted. Fixed

tomhughes:
Can you not just use `mw.loader.load(<url>)` in the `Common.js` page?

tomhughes:
See https://www.mediawiki.org/wiki/ResourceLoader/Modules#mediaWiki.loader and https://www.mediawiki.org/wiki/ResourceLoader/Legacy_JavaScript for more.

harry-wood:
oh yeah that works. DONE! https://wiki.openstreetmap.org/w/index.php?title=MediaWiki%3ACommon.js&type=revision&diff=1368345&oldid=1342124

PR #102, opened 2016-11-30
Add new blogs.openstreetmap.org site

gravitystorm:
This is the code that @tomhughes wrote, but I've changed upstream so that it generates index.html instead of planet.html. I tested it with test-kitchen and the site generates properly.


gravitystorm:
Could somebody review this PR please?

tomhughes:
I don't think there's any problem with the code (well I probably would say that since I wrote most of it ;-)) it's just there wasn't any reason to merge it until we were ready to use it - as it stands it's just an unused cookbook.

The operational question is which machine we're going to run it on...

gravitystorm:
> it's just there wasn't any reason to merge it until we were ready to use it

Fair enough. I wasn't sure from the silence whether there were problems or not or if you we reconsidering anything.

zerebubuth:
I think there's are good reasons to merge it now, even if it's currently unused:

1. Being merged into master instead of on a branch means the code won't accumulate merge conflicts that we have to fix later. I know the conflicts will exist anyway, and having it on a branch simply delays that work. However, in my experience it often builds up to levels which are onerous, and end up discouraging further progress on the PR.
2. Having lots of open PRs, especially stale ones, splits attention and conversation over many concurrent comment streams. I think it is also widely assumed that it indicates an unmaintained project.

We _plan_ to use this, so it's not unnecessary code - just unused at present. In my opinion, the best option is to merge it and deal with any further changes in a future PR, if any are necessary when we launch it.

PR #106, opened 2016-12-27
Add lldpd support

Firefishy:
Add lldp daemon to get Link Layer Discovery Protocol (LLDP) announcements from switch and send LLDP announcements to switch.

lldpcli show neighbors # show switch port + details
lldpcli show chassis # show details sent to switch


Firefishy:
Slough switches send announcements, including port.

tomhughes:
I'll have a look when I get back home later in the week...

PR #108, opened 2017-01-02
Add totp tilecache token support

Firefishy:


cookbooks/tilecache/recipes/default.rb
```diff
@@ -33,9 +33,14 @@
 package "xz-utils"
 package "openssl"
 
+# oathtool for QoS token
+package "oathtool"
+
 tilecaches = search(:node, "roles:tilecache").sort_by { |n| n[:hostname] }
 tilerenders = search(:node, "roles:tile").sort_by { |n| n[:hostname] }
 
+tilecache_tokens = data_bag_item("tilecache", "tokens")
```
tomhughes: This should be in the web bag I think as the token isn't tilecache specific. If you're ready to go live with this then I'll merge the rails side and do the chef bits to configure the key.

Firefishy: Makes sense.

tomhughes: It's in `data_bag_item("web", "passwords")` as `totp_key` now.

Firefishy:
Merged.

PR #109, opened 2017-01-23
Downloadable munin data

zerebubuth:
Add simple script to dump munin data from RRD to XML and tar it up somewhere downloadable.

Alternatively, it could copy it back to the planet server and serve it from there. I figured it was a niche enough thing that it didn't really need that, but happy to implement if the general consensus is that it should be on planet.

### DIFF
--- a/cookbooks/munin/files/default/rrddump.sh
+++ b/cookbooks/munin/files/default/rrddump.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+RRD_DIR=/var/lib/munin/openstreetmap
+DIR=`mktemp -d`
+NPROCS=8
+
+function cleanup {
+rm -rf "$DIR"
+}
+
+trap cleanup EXIT
+
+cd "$RRD_DIR"
+find -name "*.rrd" -print0 | xargs --null --max-procs=$NPROCS -I {} rrdtool dump {} "$DIR/{}.xml"
+
+cd "$DIR"
+find -name "*.xml" -print0 | tar zcf - --null -T -
--- a/cookbooks/munin/recipes/server.rb
+++ b/cookbooks/munin/recipes/server.rb
@@ -81,7 +81,13 @@ remote_directory "/srv/munin.openstreetmap.org" do
   files_owner "root"
   files_group "root"
   files_mode 0o644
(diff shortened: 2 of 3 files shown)


tomhughes:
I really don't see any reason to do this at all, just because one person wants a stick to beat us with.

zerebubuth:
I've used this data before in the past to make hand-wavy bad-statistical projections of trends. But in many cases it's better than nothing.

We already make all this data public via munin, it's just that no one really wants to write a tool to parse it all back from the images from the munin server. It remains to be seen whether anyone wants to parse the horrible XML that `rrdtool dump` outputs. Even so, I figure if it might be useful then why _not_ make it available?

I'd prefer to consider each request on its merits, and I think this one deserved the time taken to figure out how to extract the data and make it available.

Firefishy:
Would be good to publish the data.

cookbooks/munin/files/default/rrddump.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+RRD_DIR=/var/lib/munin/openstreetmap
+DIR=`mktemp -d`
+NPROCS=8
```
tomhughes: What's the advantage of this? It doesn't seem like the time taken to do the dump is important and as CPU is the most critical resource on the munin server I'd rather minimise how much the dump uses.

cookbooks/munin/recipes/server.rb
```diff
@@ -95,6 +101,21 @@
...
+  group "root"
+  mode 0o755
+end
+
+template "/etc/cron.d/rrddump" do
+  source "rrddump.cron.erb"
+  owner "root"
+  group "root"
+  mode 0o755
+end
```
tomhughes: We need something to expire old dumps or we will eventually run out of disk space.

cookbooks/munin/files/default/rrddump.sh
tomhughes: I think I might be wrong about this - there was a problem with CPU but that was because munin only really uses one CPU so it was maxed out while all the others just sat there.

Also switching to CGI mode has reduced the amount it needs on each run.

PR #110, opened 2017-01-24
Fixes to munin dump functionality

zerebubuth:
* Touch XML files to reflect date of original RRD file.
* Don't run dump in parallel.
* Only keep latest 3 files, delete older ones.

### DIFF
--- a/cookbooks/munin/templates/default/rrddump.cron.erb
+++ b/cookbooks/munin/templates/default/rrddump.cron.erb
@@ -1,3 +1,3 @@
 MAILTO=zerebubuth@gmail.com
-# do the dump in the early hours of the morning and follow, if successful, by a cleanup of the old files. i don't think these are of any historical interest, so just keep three consecutive.
-43 3 * * * www-data nice /usr/local/bin/rrddump > /srv/munin.openstreetmap.org/dumps/`date "+munin-data-%Y-%m-%d.tar.gz"` && ls -1 /srv/munin.openstreetmap.org/dumps/munin-data-*.tar.gz | sort -r | tail -n +4 | xargs rm -f
+# do the dump & cleanup in the early hours of the morning
+43 3 * * * www-data nice /usr/local/bin/rrddump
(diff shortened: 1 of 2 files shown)


cookbooks/munin/templates/default/rrddump.cron.erb
```diff
@@ -1,3 +1,3 @@
 MAILTO=zerebubuth@gmail.com
-# do the dump in the early hours of the morning
-43 3 * * * www-data nice /usr/local/bin/rrddump > /srv/munin.openstreetmap.org/dumps/`date "+munin-data-%Y-%m-%d.tar.gz"`
+# do the dump in the early hours of the morning and follow, if successful, by a cleanup of the old files. i don't think these are of any historical interest, so just keep three consecutive.
+43 3 * * * www-data nice /usr/local/bin/rrddump > /srv/munin.openstreetmap.org/dumps/`date "+munin-data-%Y-%m-%d.tar.gz"` && ls -1 /srv/munin.openstreetmap.org/dumps/munin-data-*.tar.gz | sort -r | tail -n +4 | xargs rm -f
```
Firefishy: Being pedantic, it would be nicer to have the cleanup bit in the rrddump.sh, likely run before everything else if paranoid of an unplanned exit.

tomhughes: Well that's exactly what @zerebubuth did. My comment was in a randomish location because I hadn't really thought about the best way to do it.

zerebubuth: Yeah, that was because I tried to keep `rrddump.sh` from needing to know about the names or locations of the files it was creating. That's looking like a complication now...

tomhughes: Oh I see what @Firefishy meant now. Personally I'd also use `find .. -mtime .. -delete` rather than that very complicated pipeline but it doesn't reallt matter.

Firefishy: Eeeek. Oh yeah. Not crazy about a script which outputs as tar to stdout.
More conventional would be for the script to take an input path and use that.

zerebubuth: 93209b6 changes the script to handle the whole dump & cleanup. @Firefishy, does that look better to you?

PR #113, opened 2017-02-05
Add Time to whitelist of classes to load for YAML.safe_load

zerebubuth:
The update to ruby 2.3.1 pulls in a different version of the YAML / Psych code which doesn't allow the Time class to be loaded in `safe_load`, unlike the previous version. This change whitelists Time, and appears to solve the problem. Since `state.yaml` files are not loaded from the internet, and rarely hand-edited, then this seems like an okay thing to whitelist.

This fixes an error which was preventing changesets from being replicated since 2017-02-05 11:08.

### DIFF
--- a/cookbooks/planet/files/default/replication-bin/replicate-changesets
+++ b/cookbooks/planet/files/default/replication-bin/replicate-changesets
@@ -142,7 +142,7 @@ end
 class Replicator
   def initialize(config)
     @config = YAML.safe_load(File.read(config))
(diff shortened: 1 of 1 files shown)


tomhughes:
The real problem is that I changed `load` to `safe_load` earlier to resolve rubocop warnings...

I only became aware later that this had the potential to break things, so thanks for the fix.

pnorman:
This change has been made, but the PR remains open.

PR #114, opened 2017-02-07
Delete the planet.db sqlite file on each run

harry-wood:
at least as a temporary solution to ensure spam gets blown away: https://github.com/gravitystorm/blogs.osm.org/issues/17

The sqlite file gets rebuilt completely on pluto build. Doing this on each iteration may not be what the pluto developer intended, but the result is the same ...but without the spam!


tomhughes:
It also means we're not keeping any history though - do we actually have any use or plans for that history?

Also how does this interact with the feed caching? Does that use the database or is the cache separate?

harry-wood:
Ah hmm. I had thought it was always requesting each stream anyway, but it's doing _conditional GET_ requests, which is based on the cache in the sqlite DB. So everything will be less efficient if we do this. Asking each source blog for more data. More parsing at blogs.osm.org end. 

We could do it less often, like just once per day, to get a bit of cache benefit, with daily spam flushing. Otherwise I guess we'd need to tackle the spam in the source RSS

harry-wood:
Oh we fixed this a different way

PR #115, opened 2017-02-11
Use env to start bash

pnorman:
Although it doesn't matter for OSMF servers, this file gets used as the basis for a lot of other scripts, and this is a more portable practice.

### DIFF
--- a/cookbooks/tile/templates/default/replicate.erb
+++ b/cookbooks/tile/templates/default/replicate.erb
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # DO NOT EDIT - This file is being maintained by Chef
 


tomhughes:
No! env is evil and must die! What kind of wierdass platform doesn't have it in /bin anyway?

zerebubuth:
Just out of interest, why is `env` "evil"? One of the top Google hits for "env evil" is a Guardian article titled "Stop eating Nutella and save the forests", which wasn't a lot of help :wink:

tomhughes:
Well to start with it creates potential security issues because if somebody can manage to influence the path they can cause a different program to be executed that isn't what the author expected.

pnorman:
I'm fine with not merging, but want to point out that if someone can influence the path they can probably do whatever they want anyways, including eating Nutella ;)

PR #116, opened 2017-03-12
Drop `.bm` from non-bytemark hosts.

zerebubuth:


tomhughes:
Fixed in a2461f552f8fd04afb3cf0eece8784d834199343 - needs .ic so bytemark hosts can find them.

PR #118, opened 2017-03-23
Use .osc.gz for diffs that tile replication fetches

pnorman:
This can matter on more recent versions of osm2pgsql where type detection is done by filename. The `find` cleanup line later on gets both osm and osc files, so no changes are needed there.


pnorman:
I believe 0.90.0 is the version where this matters, but am not positive.

tomhughes:
The `expire-tiles.erb` template will need fixing as well as it globs for `changes-*.osm.gz` in the expiry queue directory.

pnorman:
I changed it to use the same glob as the `find` for deleting old downloads. These could both be changed to `changes-*.osc.gz`, but matching both during the change of file name means not having to worry about missing a set of changes.

PR #119, opened 2017-04-06
Change the title to not mention 'wiki'

harry-wood:
This feeds into $wgSitename which feeds into various places https://www.mediawiki.org/wiki/Manual:$wgSitename most obviously the HTML &lt;title> tag.

@simonpoole was asking for the word 'wiki' to be removed from the HTML title, which I agree with. In fact from a site visitor's point of view, the wiki-ness of the site would ideally be largely hidden. This is one step to doing so (the other key one would be rewriting URLs)

One concern with this change would be that it might cause a rename of the "project" namespace, but from what I can tell this has taken the name "OpenStreetMap" anyway for some reason.


harry-wood:
...oh and the 'OpenStreetMap' project namespace only has one nonPage on it anyway. See here: https://wiki.osmfoundation.org/wiki/Special:AllPages?from=&to=&namespace=4

PR #126, opened 2017-08-01
Make serverinfo available on hardware.osm.org

danfai:
http://hardware.osm.org currently shows the default Ubuntu apache webpage.
With this change this domains is bound to http://hardware.openstreetmap.org as well as hardware.osm.org and SSL Certificates would be created for both.

Delete this PR if this behaviour is intended.

### DIFF
--- a/cookbooks/serverinfo/recipes/default.rb
+++ b/cookbooks/serverinfo/recipes/default.rb
@@ -67,11 +67,12 @@ execute "/srv/hardware.openstreetmap.org" do
 end
 
 ssl_certificate "hardware.openstreetmap.org" do
-  domains "hardware.openstreetmap.org"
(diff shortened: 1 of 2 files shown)


tomhughes:
It's not particularly intended or not-intended. It's just what it is and is of no real consequence one way or the other.

danfai:
I just thought it would be more convenient to write the shorter domain, just like lists.*, help.*, ...
Their behaviour is to forward first to the long version and then to the https one (leaving the short one with SSL not possible to reach, but this is another issue).

JamesKingdom:
This was done in 26eeab924cfb9f5f9289b5dcc1860d9bc9afe6ff , so this can be closed.

Firefishy:
Sorry @danfai had not seen this PR when I made the fix.

PR #135, opened 2017-11-13
Add mcelog daemon

Firefishy:
Add the mcelog daemon to log machine check events.

The Corrected Error (CE) ECC events are useful. The daemon by default attempts to "Offline" memory which has repeated CE errors.

We should likely add some custom alerting via `/etc/mcelog/[error type]-trigger.local` scripts

### DIFF
--- a/cookbooks/hardware/recipes/default.rb
+++ b/cookbooks/hardware/recipes/default.rb
@@ -200,6 +200,13 @@ service "lldpd" do
   supports :status => true, :restart => true, :reload => true
 end
 
+# mcelog Daemon to log / alert on machine check events
+package "mcelog"
+service "mcelog" do
+  action [:start, :enable]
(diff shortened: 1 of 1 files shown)


tas50:
I'm not sure how OSM feels about external deps, but I wrote this cookbook at a previous job and we're managing it at Chef now https://github.com/chef-cookbooks/mcelog

Fully managing the mcelog config is useful so you make sure you get the CPU failure logging as well as the DIMM logging. We founds tons of stuff over the year we ran it.

tomhughes:
We're not really setup for using external cookbooks at the moment - our setup is rather old and other than a few things that we just copied in early on everything has been written from scratch by us.

tas50:
@tomhughes You ever want some help with Chef things I'm always available. I'm on the Chef community slack with this same handle

PR #141, opened 2018-01-14
add skin to OSMF mediawiki instance

nebulon42:
Trying to refresh the mediawiki skin on the OSMF mediawiki instance. The repository for the skin lives at https://github.com/osmfoundation/osmf-mediawiki-skin. This PR is meant to enable individual users on that instance to be able to test the skin without making it the default.

Please note that my Ruby knowledge is only basic and my Chef knowledge is nearly nonexistent.

### DIFF
--- a/cookbooks/foundation/recipes/wiki.rb
+++ b/cookbooks/foundation/recipes/wiki.rb
@@ -48,6 +48,13 @@ mediawiki_skin "osmf" do
   revision "master"
 end
 
+mediawiki_skin "OSMFoundation" do
+  site "wiki.osmfoundation.org"
+  repository "git://github.com/osmfoundation/osmf-mediawiki-skin.git"
+  revision "master"
+  legacy False
+end
+
 cookbook_file "/srv/wiki.osmfoundation.org/Wiki.png" do
   owner node[:mediawiki][:user]
   group node[:mediawiki][:group]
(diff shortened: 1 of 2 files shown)


tomhughes:
Can you explain what the `legacy` stuff is all about and why we need it?

nebulon42:
Ah, sorry, forgot about that. Prior to Mediawiki 1.24 you used `include_once` to enable a skin. After that the preferred method is to use `wfLoadSkin` (see https://www.mediawiki.org/wiki/Manual:Skin_configuration#Enabling_skins). But the skin must support that by providing a `skin.json` file.

Since the current skin for the OSMF wiki doesn't support the new method I thought we need some extension to skin loading without the need to touch all the other places where the recipe might be used. Hence the introduction of the `legacy` parameter which defaults to true and can be omitted in the scenarios that already use the deprecated way of enabling a skin.

But if you can think of some better method, and I guess you can, then I'm happy to change that.

harry-wood:
CWG have been discussing this with @nebulon42 , and we are in favour of deploying this. We'll test it and pretty soon I'm sure we can set it as the default skin and blow away the current skin on there.

The current skin is a slightly weird font and clumsy design in various ways. I created it to match a Wordpress instance we used to have running alongside. This looks like a nice improvement

nebulon42:
Thanks!

PR #143, opened 2018-01-16
Use https URL for html base tag for nominatim.openstreetmap.org.

grischard:
The <base> tag on nominatim.openstreetmap.org currently keeps the protocol. This changes it to always direct browsers to the https version, encouraging https without the breakage risks that a redirect would bring.


lonvia:
Merged as c0d36445d65eacdd7dc9b3cae88ab5b5f59c842c.

PR #145, opened 2018-01-24
Add cron job to delete old planet files.

zerebubuth:
At the moment it's just hard-coded to the general consensus view in https://github.com/openstreetmap/operations/issues/180 that we should keep the last 4 weeks of planet files, and the first planet for each calendar month before that.

This doesn't touch any of the "historical interest" planets in `/store/planet/cc-by-sa`.

Let me know if there are other dates that we have special reasons for wanting to keep, and I'll whitelist them.

### DIFF
--- a/cookbooks/planet/templates/default/old-planet-file-cleanup.erb
+++ b/cookbooks/planet/templates/default/old-planet-file-cleanup.erb
@@ -1,0 +1,124 @@
+#!/usr/bin/ruby
+
+require 'date'
+require 'optparse'
+
+# always keep the last 4 weeks
+ALWAYS_KEEP_DAYS = 4 * 7
+
+# otherwise, bucket by month and keep the earliest in the bucket
+def bucket(date)
+  Date.new(date.year, date.month, 1)
+end
+
+Candidate = Struct.new(:filename, :date)
+
+def list_files(glob, date_pattern)
+  # find all candidates for deletion
+  real_files = Dir.glob(glob).select do |file|
+    File.file?(file) && !File.symlink?(file)
+  end
+
+  real_files.map do |file|
+    # extract the date
+    m = date_pattern.match(file)
+    raise "Unable to extract date string from #{file.inspect}" if m.nil?
+    d = Date.strptime(m[1], "%y%m%d")
+    Candidate.new(file, d)
+  end
+end
+
+def deletion_candidates(today, candidates)
+  candidate_buckets = Hash.new
+
+  candidates.each do |c|
+    next if today - c.date < ALWAYS_KEEP_DAYS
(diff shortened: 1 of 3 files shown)


zerebubuth:
Ooops, I should have noted that it's configured to run in "dry run" mode at the moment, and just email me the list of files that it would delete. I'll let that run for a couple of months, just to check it's doing the right thing, and follow up with PR to enable it properly.

cookbooks/planet/templates/default/old-planet-file-cleanup.erb
```diff
@@ -0,0 +1,124 @@
...
+
+to_delete += deletions(
+  "#{xml_directory}/20??/planet-??????.osm.bz2",
+  /planet-([0-9]{6}).osm.bz2/,
+  today,
+  ["changesets-%y%m%d.osm.bz2",
+   "changesets-%y%m%d.osm.bz2.md5",
+   "discussions-%y%m%d.osm.bz2",
+   "discussions-%y%m%d.osm.bz2.md5",
+   "planet-%y%m%d.osm.bz2",
```
Firefishy: I'd prefer keeping at least the osm.bz2 file around and blowing away the PBF etc copies.

The planet osm.bz2 files are universally understood because it is a simple plain text format, now and into the future. The history osm.bz2 files have a higher barrier to entry.

tomhughes:
Pretty sure that ship is long sailed @Firefishy as I've been blowing shit away at will for a year or so now in a desparate attempt to stop ironbelly sinking beneath the waves.

I suspect my "expiry policy" has been less generous that @zerebubuth as well. Didn't I only keep one file a year for at least some things?

zerebubuth:
I've been deleting stuff on an ad-hoc basis as well. This is an attempt to make the process more predictable and automatic.

@Firefishy, I know the old planets are tempting to keep around, but we face 2 problems:

1. Quadratically increasing disk space usage over time. Not only do we run out of disk, but we have to manage that capacity across any redundant sites.
2. Distributing old planet files which may contain data we're no longer allowed to distribute (i.e: redacted for copyright or "right to be forgotten" reasons).

The consensus in https://github.com/openstreetmap/operations/issues/180 seemed to be that people still wanted _some_ old planets, but once a month was enough to make point-in-time feasible by applying diffs, while still freeing up a comparatively large amount of space.

After getting this temporary fix implemented, I'd like to address the main concern by making a fast point-in-time planet download service. Then we can consider ceasing to distribute old, problematic planets.

And after that, we need to figure out what we're going to do about diffs.

tomhughes:
Merged. A dry run says:

```
Would delete files of total size 9873.284852423 GB
Would delete 1016 files
```

;-)

PR #151, opened 2018-02-17
Add MIME type for .osm uploads

h3lix1:
Continuation of fix for operations/194

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -104,7 +104,7 @@ $wgFileExtensions[] = 'osm';
 # for .osm files.
 $wgXMLMimeTypes[] = array('osm' => 'application/vnd.openstreetmap.data+xml');
 
-$wgTrustedMediaFormats[] = 'application/vnd.openstreetmap.data+xml'
+$wgTrustedMediaFormats[] = 'application/vnd.openstreetmap.data+xml';
 
 $wgSVGConverters = array( 'rsvg' => '/usr/bin/rsvg-convert -w $width -h $height -o $output $input');
 $wgSVGConverter = 'rsvg';
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -98,6 +98,20 @@ $wgFileExtensions[] = 'osm';
 	$wgFileExtensions[] = '<%= mw_extra_file_extension %>';
 <% end -%>
 
+# Add OSM XML file format per http://www.iana.org/assignments/media-types/media-types.xhtml
+# Shout out to Paul Norman for reserving this.
+$wgXMLMimeTypes[] = 'osm' => 'application/vnd.openstreetmap.data+xml';
+
+# Helps MimeMagic determine XML-based formats and chooses the correct MimeType
+# for .osm files.
+$wgXMLMimeTypes = array(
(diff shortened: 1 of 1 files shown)


cookbooks/mediawiki/templates/default/LocalSettings.php.erb
```diff
@@ -98,6 +98,23 @@ $wgFileExtensions[] = 'osm';
 	$wgFileExtensions[] = '<%= mw_extra_file_extension %>';
 <% end -%>
 
+# Add OSM XML file format per http://www.iana.org/assignments/media-types/media-types.xhtml
+# Shout out to Paul Norman for reserving this.
+$wgXMLMimeTypes[] = 'osm' => 'application/vnd.openstreetmap.data+xml';
```
tomhughes: Sorry but I don't know PHP at all, so can you explain what this (which looks a little odd) actually does? Am I right in assuming that this adds a new entry to an existing hash?

cookbooks/mediawiki/templates/default/LocalSettings.php.erb
```diff
@@ -98,6 +98,23 @@ $wgFileExtensions[] = 'osm';
...
+
+# Helps MimeMagic determine XML-based formats and chooses the correct MimeType
+# for .osm files.
+$wgXMLMimeTypes = array(
+	'http://www.w3.org/2000/svg:svg' => 'image/svg+xml',
+	'svg' => 'image/svg+xml',
+	'http://www.lysator.liu.se/~alla/dia/:diagram' => 'application/x-dia-diagram',
+	'http://www.w3.org/1999/xhtml:html' => 'text/html',
+	'html' => 'text/html',
+	'osm' => 'application/vnd.openstreetmap.data+xml' );
```
tomhughes: What's the logic here? It looks like this is doing a complete replacement of the hash which is not really ideal as it means we won't get any future updates to the default value, and in any case why would we add a new entry and then replace the whole thing?

cookbooks/mediawiki/templates/default/LocalSettings.php.erb
```diff
@@ -98,6 +98,23 @@ $wgFileExtensions[] = 'osm';
...
+
+$wgTrustedMediaFormats[] = 'application/xml'
+$wgTrustedMediaFormats[] = 'application/vnd.openstreetmap.data+xml'
+
```
tomhughes: Why do we need to add `application/xml` as well as the OpenStreetMap one?

cookbooks/mediawiki/templates/default/LocalSettings.php.erb
h3lix1: This line is actually incorrect - you are right for bringing this up.

I meant to just submit the configuration below instead. Although you mention keeping this dynamic, so I will re-visit this. (and correct it)

cookbooks/mediawiki/templates/default/LocalSettings.php.erb
h3lix1: This is correct - it is replacing the array. I will remove this and fix the above entry for $wgXMLMimeTypes to add the elements to the array dynamically.

cookbooks/mediawiki/templates/default/LocalSettings.php.erb
h3lix1: In theory, the second config value should work perfectly well, but I can't test it to verify. This whitelists both formats in case it doesn't work as expected. (covering bases) I'll add only one to reduce the cruft.

h3lix1:
Made changes as requested to make dynamic, and to remove the extra whitelist entry that shouldn't be necessary.

PR #152, opened 2018-02-20
Set autovacuum scale factors for tile servers

pnorman:
Reducing these factors helps make sure autovacuum runs often enough and keeps bloat down.

I haven't tested this code change, but have tested these values on render servers and find they work far better than the defaults.

This change will **not** remove existing bloat but cap it's future growth.


tomhughes:
Merged, after first adding 27e33fa61307e73ca2b81bc16ac09699878970ce to make those configurable ;-)

PR #157, opened 2018-04-16
Enable old planet file cleanup

zerebubuth:
Enable the old planet file cleanup on ironbelly and grisu by removing the `--dry-run` flag from the script.

It has been running as expected, except for a typo which meant it wasn't cleaning up old XML full history planet files. That's fixed below.

If you're happy with this, and we merge it, then I'll run it by hand once and again in `--dry-run`, just to make doubly sure that it's doing what it's supposed to in-situ. (I already tested this locally, but always worth being careful!)


zerebubuth:
FYI - I ran the script. It cleaned up about 10TB of space, taking `/store` usage down to 64%. A second (dry) run identified no files to clean up. So it appears to be working. I'll keep an eye on it for the next few cron runs.

PR #160, opened 2018-04-30
Remove MobileFrontend Mediawiki extension from OSMF wiki

### LINKED ISSUE #159: Disable MobileFrontend at OSMF wiki
I'd like to disable the MobileFrontend extension at OSMF wiki while keeping it active at the OSM wiki. The reason is that the new OSMF wiki theme is responsive and so the dedicated mobile frontend should not being displayed.

With my limited knowledge of Chef I figured out that all the extensions live in inc.php files that are somehow all included in the LocalConfig. How would we specify which extensions we want to have enabled? Write a list and only include those in the list in the LocalConfig?

Ref https://github.com/osmfoundation/osmf-mediawiki-skin/issues/9

nebulon42:
Fixes #159

### DIFF
--- a/cookbooks/foundation/recipes/board.rb
+++ b/cookbooks/foundation/recipes/board.rb
@@ -43,3 +43,9 @@ cookbook_file "/srv/board.osmfoundation.org/Wiki.png" do
   group node[:mediawiki][:group]
   mode 0o644
 end
+
+mediawiki_extension "MobileFrontend" do
+  site "board.osmfoundation.org"
+  template "mw-ext-MobileFrontend.inc.php.erb"
+  update_site false
+end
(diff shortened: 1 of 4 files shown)


cookbooks/foundation/recipes/board.rb
```diff
@@ -43,3 +43,9 @@
   group node[:mediawiki][:group]
   mode 0o644
 end
+
+mediawiki_extension "MobileFrontend" do
+  site new_resource.site
```
tomhughes: That will need to be the name of the site (ie `"board.osmfoundation.org"`) not `new_resource.site` which only makes sense in the context of the resource where it was before.

nebulon42: Thanks I missed that. Updated.

PR #164, opened 2018-06-01
Redirect tilecache http requests to https

grischard:
* Make squid listen on 127.0.0.1:8080
* Point nginx to 127.0.0.1:8080
* Make nginx listen on port 80 and redirect http requests there to https

Closes https://github.com/openstreetmap/operations/issues/117 and https://github.com/openstreetmap/operations/issues/190.

### DIFF
--- a/cookbooks/squid/templates/default/squid.conf.erb
+++ b/cookbooks/squid/templates/default/squid.conf.erb
@@ -12,7 +12,7 @@ icp_port 3130
 log_icp_queries off
 
 #FIXME - configurable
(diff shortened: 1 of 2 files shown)


tomhughes:
What's the impact of this on the rate limiting @Firefishy? Am I right in thinking that the rate limiting works differently (and may be less effective) for traffic that goes via nginx?

grischard:
After our IRC discussions: should we merge this as it is, as the effect will be minimal? Is it worth rewriting the squid config if we'll need to go back to nginx anyway the day we get around to configuring varnish?

tomhughes:
Merged (with fixes) as 981b586040b3fe307a4855e58e846da75555aa05.

PR #165, opened 2018-06-03
Allow hsts preloading

grischard:
Merging this will allow [hsts preloading](https://hstspreload.org). There haven't been any complaints about our global hsts switch so far, so I'm pretty sure this is ok, but I'd still wait a week or so as there's no undo button on this. 

Contributes to https://github.com/openstreetmap/operations/issues/190 and https://github.com/openstreetmap/operations/issues/117


tomhughes:
Waiting a week before merging this? or before applying for preload?

grischard:
Before merging, because IIRC anyone can apply to preloading for us then.

grischard:
@tomhughes let's do this?

PR #183, opened 2018-08-25
Recipe to keep a local updated planet file.

joto:
When we create coastline files on OSMF servers in the future we'll need an updated planet file. This recipe is a building block for this, creating a role for keeping a planet up-to-date. The `taginfo` role can already be changed to use this as well as the coastline code (for which I will supply a pull request at a later date). This has a script to update the planet file and a wrapper script that will call the update script and then call all scripts it finds in `/srv/cplanet/jobs` (each run as the owner of the script), similar to how cron spool files work. This way taginfo, coastline update etc. can all hook into this and run their processing every time a new planet is available.

@tomhughes


joto:
Btw this needs a new user `cplanet`.

cookbooks/cplanet/templates/default/update.erb
```diff
@@ -0,0 +1,30 @@
...
+    for job in `find "$DIR/jobs" -type f -executable | sort`; do
+        user=`stat '--format=%U' $job`
+        echo "Running '$job' as user '$user'..."
+        su -c "$job" "$user"
+    done
+fi
+
+echo
+echo "Expire old logs..."
+find "${LOGDIR}" -mtime +28 -delete
```
Firefishy: Paranoid option would be: ```find "${LOGDIR}" -mtime +28 -type f -iname "*.log" -delete```

cookbooks/cplanet/templates/default/update-planet.erb
```diff
@@ -0,0 +1,26 @@
...
+PLANETDIR="<%= @basedir %>/planet"
+PLANETPREV="${PLANETDIR}/planet-previous.pbf"
+PLANETCURR="${PLANETDIR}/planet.pbf"
+PLANETNEW="${PLANETDIR}/planet-new.pbf"
+
+retval=1
+while [ $retval -eq 1 ]; do
+    pyosmium-up-to-date -v -o $PLANETNEW $PLANETCURR
+    retval=$?
+done
```
tomhughes: So I discovered an interesting issue with this loop today while spinning up a new taginfo (which uses the same logic) from scratch.

The problem is that if planet is old enough that there is more than 1Gb of changes then `pyosmium-up-to-date` will stop after 1Gb but will exit one causing us to go round the loop and try again - except that `planet-new` hasn't been renamed over `planet` to it just tries to do the same thing again, but fails as `planet-new` already exists...

PR #185, opened 2018-09-05
Add a script to generate a list of deleted users uids

simonpoole:
No idea if this works and there don't seem to be tests for anything similar. Lets see what CI says.

### DIFF
--- a/cookbooks/planet/templates/default/users-deleted.erb
+++ b/cookbooks/planet/templates/default/users-deleted.erb
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+# DO NOT EDIT - This file is being maintained by Chef
+
+T=$(mktemp -d -t -p /var/tmp users.XXXXXXXXXX)
+
+# use the same as for the users-agreed template
+export PGPASSFILE=/etc/replication/users-agreed.conf
+
+echo "# user IDs of deleted users. " > $T/users_deleted
+psql -h <%= node[:web][:readonly_database_host] %> -U planetdiff -t -c "select id from users where status='deleted' by id asc" openstreetmap >> $T/users_deleted
+
+if cmp -s "${T}/users_deleted "/store/planet/users_deleted.txt"; then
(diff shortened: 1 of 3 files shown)


cookbooks/planet/templates/default/users-deleted.erb
```diff
@@ -0,0 +1,19 @@
...
+
+# DO NOT EDIT - This file is being maintained by Chef
+
+T=$(mktemp -d -t -p /var/tmp users.XXXXXXXXXX)
+
+# use the same as for the users-agreed template
+export PGPASSFILE=/etc/replication/users-agreed.conf
+
+echo "# user IDs of deleted users. " > $T/users_deleted
+psql -h <%= node[:web][:readonly_database_host] %> -U planetdiff -t -c "select id from users where status='deleted' by id asc" openstreetmap >> $T/users_deleted
```
tomhughes: I think you meant "order by id asc" here.

simonpoole: yes ... copy paste error ...

mmd-osm:
@simonpoole : I believe this change was somehow GDPR related, but I couldn't figure out why the SELECT statement didn't include `changesets_count > 0` as a condition. It would be a prerequisite for a user to show up in any public facing planet dump or diff file.

Nowadays, the users_deleted.txt file is fairly large, and most users are likely spam accounts that never contributed anything.

simonpoole:
While having a non-zero number of changesets is a prerequisite for tuning up in the data dumps, it is conceivable that the user in question used the account to authenticate in a 3rd party service, so I'm not quite sure that they could just be left away.

mmd-osm:
> maybe we should simply have two lists

Yes, that sounds like a good idea.

mmd-osm:
Any idea what would be a good filename for the second list? Something along the lines of _active_users_deleted,_ or  _contributors_deleted_, maybe?

PR #187, opened 2018-09-07
Tlie expiry on bionic still needs ruby

jburgess777:
Currently the tile expiry on pyrene is failing because ruby isn't installed:
```
$ /usr/local/bin/expire-tiles
-bash: /usr/local/bin/expire-tiles: /usr/bin/ruby: bad interpreter: No such file or directory
```

### DIFF
--- a/cookbooks/tile/recipes/default.rb
+++ b/cookbooks/tile/recipes/default.rb
@@ -440,6 +440,7 @@ if node[:lsb][:release].to_f >= 18.04
   package %w[
     pyosmium
     python-pyproj
+    ruby
   ]
 
   remote_directory "/usr/local/bin" do


jburgess777:
After pyrene picks up the change we will probably need to force any existing tiles to be rendered again in case they missed a change:
```
sudo touch /srv/tile.openstreetmap.org/tiles/default/planet-import-complete
```

jburgess777:
I see @Firefishy just committed an equivalent fix

Firefishy:
Whoops, only saw this now.

tomhughes:
I have reverted the touch for now - a low zoom render needs to be done first or the touch will cause all the low zoom tiles to be treated as dirty.

tomhughes:
In fact I don't think it was necessary to play with that at all - the pending expiries should still have been queued and would have been processed once ruby was installed.

PR #188, opened 2018-09-10
Scripts for "export" server generating coastline and icesheet data.

joto:


joto:
This needs the PR #183 and a user called `export`.

@tomhughes

joto:
Some more details:
* Currently this doesn't hook into the `cplanet` generation yet. Can be added easily once `cplanet` is available.
* The `update` script generates the coastline and icesheet data. Generates files will end up in `unchecked` directory and hardlinked to `checked` directory if the check comparing the newly rendered image of the coastlines is similar enough to the last known good image. From there the data still has to be copied somewhere, for instance the `planet.osm.org` server.
* It also generates the files needed for the OSM Inspector coastline view. There is no provision yet to copy this data to the actual OSM Inspector server.
* We still have to figure out how to handle the case when the coastline "is stuck" because it is too different from last known good version. Currently to get it unstuck, the `master.tiff` link needs to be removed and the next run will go through.

I believe we should deploy this now and test it. (I have done some testing but could not do it with planet file due to limitations in my testing VM.) Once the basic stuff all works we can figure ot the rest.

tomhughes:
Well I need to sort out cplanet before we can deploy this!

This will likely need some work as well - if nothing else it needs a better name that `export` although I don't know what yet.

joto:
I'll do the coastline/icesheet processing somewhere else.

tomhughes:
This was literally still queued in my inbox to work on - just needed to actually find a machine to use.

I'm happy to continue to pursue it if you want.

PR #198, opened 2018-10-03
Attempt to make changeset replication more robust

zerebubuth:
Attempt to make changeset replication more robust to overruns and concurrent executions by not copying the state file, instead overwrite it in-place. Also, try to make the recovery process more robust if a zero-length state file is found.

### DIFF
--- a/cookbooks/planet/files/default/replication-bin/replicate-changesets
+++ b/cookbooks/planet/files/default/replication-bin/replicate-changesets
@@ -142,10 +142,6 @@ end
 class Replicator
   def initialize(config)
     @config = YAML.safe_load(File.read(config))
-    @state = YAML.safe_load(File.read(@config["state_file"]), [Time])
-    @conn = PG::Connection.connect(@config["db"])
-    # get current time from the database rather than the current system
-    @now = @conn.exec("select now() as now").map { |row| Time.parse(row["now"]) }[0]
   end
 
   def open_changesets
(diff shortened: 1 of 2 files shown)


zerebubuth:
> Yowsers. That's, err, complicated....

Yup! Definitely easier to read the [first version of the PR](https://github.com/openstreetmap/chef/pull/198/commits/5096123ef41c5e1806c1583deaa00a15803ff2af) before I had to move a bunch of stuff around to satisfy rubocop!

> Maybe we should just use a separate lock file rather than trying to use flock on the state file?

Yup, could do that instead. I thought it was a little inelegant to have a separate file... but it turns out my alternative single-file implementation is pretty inelegant too.

cookbooks/planet/files/default/replication-bin/replicate-changesets
```diff
@@ -207,63 +277,21 @@ class Replicator
...
+
+    File.open(lock_file, File::RDWR | File::CREAT) do |fl|
       fl.flock(File::LOCK_EX)
 
+      @state = read_state(state_file)
+
+      # initialise this stuff _after_ we've got the lock on the state file.
+      @conn = PG::Connection.connect(@config["db"])
+      # get current time from the database rather than the current system
+      @now = @conn.exec("select now() as now").map { |row| Time.parse(row["now"]) }[0]
```
mmd-osm: If _making replication more robust_ includes dealing with a non-utc server tz, something like `now() at time zone 'utc'`  might be helpful here.

mmd-osm:
Wasn't this changed to 10'000 changes some time ago?

```
# after this many changes, a changeset will be closed
CHANGES_LIMIT = 50000
```

pnorman:
> Wasn't this changed to 10'000 changes some time ago?

Yes.

mmd-osm:
Some users reported loss of data on 2019-06-15 19:43  - https://planet.openstreetmap.org/replication/changesets/003/457/

https://mobile.twitter.com/pascal_n/status/1144505676863086593

mmd-osm:
This replication broke again yesterday on almost the same time as last year: https://mobile.twitter.com/pascal_n/status/1250856509011042306

zerebubuth:
Interesting. Thanks for pointing that out. As far as I can figure out, there's no time dependency in the code, so the breakage anniversary might just be a coincidence. But it's possible that the probability of breakage is increased by other things happening on the machine; load, network congestion, etc...

Apologies to anyone impacted by the breakage. I just fixed it, and a longer term fix (i.e: this PR) is still on my TODO list (after 1.5 years... :roll_eyes: ). If anyone would like to help out, that would be gratefully received. The minimum viable fix (which might not be this PR) is to:

1. Don't lock on the state file, lock on some other file in `/var/run/replication/changeset` or something like that. The issue at its core is that the script locks the state file, but then moves the new state file over the top of it, which leads to filesystem races. (There's a good reason to atomically replace the state file, I just made a mistake combining that with locking.)
2. If the lock can't be acquired, exit and don't block. The script currently blocks, which means when it takes N minutes to run the script there can be N-1 processes all racing to replace the file afterwards.

mgd722:
@zerebubuth Though it seems to be running well now, I'm not sure that the replication caught up correctly. Based on the absence of replication files, it seems the outage was from `2020-04-16 17:41` to `2020-04-17 08:31`. With a fresh copy of ChangesetMD, I began a new replication starting from `2020-04-16 12:59:00` (sequence `3/895/206`) and ran it through now. In doing so I should have captured changeset [83667994](https://www.openstreetmap.org/changeset/83667994#map=15/1.6550/103.5688) which took place at `2020-04-16 20:57:42`, but it's not in the database. The ["catch-up" replication file](https://planet.osm.org/replication/changesets/003/895/489.osm.gz), which should contain all changesets that took place during the outage, does not have this changeset in it. Additionally, just based on scrolling through what did get replicated in the catch-up file, it seems it is missing many changesets.

zerebubuth:
@mgd722 thanks for the investigation!

I think what's happening is that the catchup process is broken, probably because the code isn't designed to handle such a long catchup. (And it has probably never been tested for it, either.) Sorry about that! I'm not sure that there's anything I can do to repair the files at this point without causing more problems by "rewriting history" for anyone who has already downloaded the files... :disappointed:

mmd-osm:
We’ve been discussing this on slack that backfilling via the forthcoming weekly dump should be feasible. It just takes some more waiting time for the dump to be available.

zerebubuth:
Yes, I agree it's feasible. If anyone is going to put development effort into figuring it out, I'd beg them to put it into fixing this replication process instead.

Once the full dump is available, how useful are the individual replication diffs? Could we simply delete them, given all the same information is available in the full dump?

mgd722:
At least for my purposes, backfilling my database from the weekly dump is pretty straightforward. The timing was quite unfortunate in that the last weekly dump was 31 minutes before the outage! I'm a full stack engineer, but taking a look so far this seems way deeper into devops than I really understand. I'll read up a bit more to see if I can follow what's going on.

Regarding keeping old minutely replication files, I would think that most could be deleted. Maybe just save 90 days worth? That way in case your replication falls behind for a over a week you don't need to load all 84 million changesets to fill in the gap. I'm definitely not familiar enough with the rest of the OSM ecosystem to say if anything else depends on minutely replication though.

I know this is more of a bandaid than anything, but shortly I'll be building a service to monitor the minutely dumps so next time I'll be able to report it a bit sooner.

PR #209, opened 2018-12-14
Add firewall_rule helper support

Firefishy:
Adds helper support to firewall_rule.

Enable netfilter FTP helper for ftp cookbook.

### DIFF
--- a/cookbooks/ftp/recipes/default.rb
+++ b/cookbooks/ftp/recipes/default.rb
@@ -50,4 +50,5 @@ firewall_rule "accept-ftp-tcp" do
   proto "tcp"
   dest_ports "ftp"
   source_ports "-"
+  helper "ftp"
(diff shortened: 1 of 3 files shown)


cookbooks/ftp/recipes/default.rb
```diff
@@ -51,3 +51,14 @@
   dest_ports "ftp"
   source_ports "-"
 end
+
+# Enable Netfilter ftp helper on port 21
+firewall_rule "helper-ftp-tcp" do
```
tomhughes: I don't think this needs (or wants) an extra `HELPER` rule - you can just specify the helper on the existing `ACCEPT` rule by my reading of the manual page.

The ability to use a `HELPER` rule is just there for cases where the packets are already being accepted by the default policy.

PR #214, opened 2019-01-24
add Thanks extension to Wiki (incl. Echo extension as dependency)

Tigerfell:
This is an attempt to fix openstreetmap/operations#265. This should install the MediaWiki extensions "Echo" and "Thanks" to the wikis. 
AFAIK we need version 1.31 of "Echo" because the current master [requires MediaWiki >= 1.32.0](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Echo/+/master/extension.json#18) and we just run [1.31.1](https://wiki.openstreetmap.org/wiki/Special:Version). This can be obtained from MediaWiki itself: https://www.mediawiki.org/wiki/Special:ExtensionDistributor/Thanks (possibly also from their repos).
Please let me know if there is anything wrong with this PR. I will try to fix it.


tomhughes:
Well that won't do anything - that just adds templates for the configuration but it never actually uses them and certainly doesn't install the extensions.

Firstly those should probably be in the wiki cookbook (unless we're going to install this on the other board and DWG wikis as well) and then you need something like https://github.com/openstreetmap/chef/blob/master/cookbooks/wiki/recipes/default.rb#L59 to actually make use of them.

Note that this will only work if they can be installed from the mediawiki github, and if they need to be composed then you need to make sure you enable that when declaring the resource.

tomhughes:
There shouldn't be any reason to specify a version - by default we will track the REL1_31 branch in the github repo (or equivalent based on the current MW version).

Tigerfell:
Great, I updated and squashed it. Anything else that needs to be changed?

Tigerfell:
I was not really sure about it, so I added it just in case.
Thanks again!

PR #220, opened 2019-02-23
Wiki Dumps

thomersch:
As reported in https://github.com/openstreetmap/operations/issues/179, the wiki dumps are broken for some time. This PR attempts to fix the process.

A change in the DNS would be necessary as dump.wiki.openstreetmap.org does not resolve anymore. I can provide you a patch for this as well.

(This is my first time working with Chef, so I am happy to learn)


tomhughes:
I don't think the breakage was accidental. I think they were dropped for being too resource intensive.

Firefishy:
I honestly cannot remember why they were disabled. I think it was likely an accident during transition, or a bit that was never completed in the move to chef.

tomhughes:
Well whatever... As far as I can see the cron job in the repository (even going back into the private repo) never did anything.

I'm not really sure what use the dump is anyway? What would anybody want it for?

As far as this PR goes it definitely needs to upgrade the apache configuration with SSL support and to add the dump name to the certificate.

Does it actually need a separate site though? or could we just serve the dump from wiki.openstreetmap.org?

thomersch:
I am doing this for @raubraupe who is working on cleaning up the wiki and who wants to generate statistical data about languages/communities/types of content and doesn't want to hammer the API.

I would be ok with changing the configuration to use a subpath instead of a subdomain, I just wanted to do a minimal change to what I thought was the previous state.

PR #224, opened 2019-02-24
Install specific commit of civiruleshttppost and latest on branch of osm username

grischard:


tomhughes:
If you really want to track a branch you should name the branch - firstly because I'm not sure what `revision nil` will actually do, and we always pass the revision to the `git` resource and secondly because the documentation says he default is `HEAD` but that is meaningless unless you're already in a checkout and even if you were it would mean the revision you currently had checked out and hence do nothing...

The bigger worry is that something without releases is probably not being well maintained - indeed it appears that `civiruleshttppost` has had no commits for four years.

PR #232, opened 2019-04-23
Allow sysops to modify interwiki table

Tigerfell:
Assigns ```interwiki``` permission to sysops, so they can modify the interwiki table. 
I suggested this change when we were discussing about renaming the project namespace to ```Wiki```. I think removing the interwiki link with this prefix is a prerequisite. 

Link to the discussion: https://wiki.openstreetmap.org/w/index.php?title=Talk:Wiki&diff=prev&oldid=1844371
Link to MediaWiki docs: https://www.mediawiki.org/wiki/Extension:Interwiki#Installation

Closes openstreetmap/operations#296.

### DIFF
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -262,6 +262,7 @@ action :create do
 
   mediawiki_extension "Interwiki" do
     site new_resource.site
+    template "mw-ext-Interwiki.inc.php.erb"
     update_site false
   end
 
--- a/cookbooks/mediawiki/templates/default/mw-ext-Interwiki.inc.php.erb
+++ b/cookbooks/mediawiki/templates/default/mw-ext-Interwiki.inc.php.erb
@@ -1,0 +1,5 @@
+<?php
+# DO NOT EDIT - This file is being maintained by Chef
+
+# allow sysops to modify interwiki table
+$wgGroupPermissions['sysop']['interwiki'] = true;


tomhughes:
Please fix the indentation errors reported (don't worry about the block length).

Also please clearly specify if there are manual steps that need be taken before this change is made and detail exactly what needs to be done (at least I assume your comment about a prerequisite means there is something that needs to be done).

Tigerfell:
After this change, a wiki admin would need to remove the entry ```Wiki``` from the configuration page [Special:Interwiki](https://wiki.openstreetmap.org/wiki/Special:Interwiki). Then, the following line would need to be updated https://github.com/openstreetmap/chef/blob/f6059322a706366a4dfd2e0d8d582d63f2a97e91/cookbooks/wiki/recipes/default.rb#L50

to ```metanamespace "Wiki"```. 

I guess Travis now complains about the coding style?

Tigerfell:
@Firefishy What do you mean by require a rebuild?
This PR just assigns a permission to sysops.

Firefishy:
I am concerned that changing `metanamespace "OpenStreetMap"` may require a search or similar rebuild and manual work. The code used here is used for more than just wiki.openstreetmap.org, it is also used for the wiki.osmfoundation.org and a private project wiki.

If you are confident that the change is only a simple config change, then OK.

Tigerfell:
I will open a separate PR for that change which will modify the ```wiki``` cookbook only. So, this is just a config change. 
PS: I just forced the two commits into one.

Tigerfell:
@Firefishy Is there anything left to do for me?

Tigerfell:
Just for the record someone already set up a wiki page for requesting changes to the interwiki table, so it would be really helpful if the admins could respond to that. 
https://wiki.openstreetmap.org/wiki/Interwiki

tomhughes:
Really I think that's something the wiki admins should deal with - if they need us to implement the decision then fine.

PR #234, opened 2019-04-28
Rename meta namespace in main wiki

Tigerfell:
The current name of the project/meta namespace was considered confusing, because the pages in this namespace contain organisational content about the wiki (see https://wiki.openstreetmap.org/wiki/OpenStreetMap:VisualEditor for instance), the name ```OpenStreetMap``` implied however some relevance for OSM itself. 

Renaming was discussed at https://wiki.openstreetmap.org/w/index.php?title=Talk:Wiki&oldid=1846962#Renaming_the_project_name_space
I already carried out some manual fixing to avoid breaking links and I will check again after renaming.

### DIFF
--- a/cookbooks/wiki/recipes/default.rb
+++ b/cookbooks/wiki/recipes/default.rb
@@ -47,7 +47,7 @@ mediawiki_site "wiki.openstreetmap.org" do
   email_sender "wiki@noreply.openstreetmap.org"
   email_sender_name "OpenStreetMap Wiki"
 
-  metanamespace "OpenStreetMap"
+  metanamespace "Wiki"
 
   recaptcha_public_key "6LdFIQATAAAAAMwtHeI8KDgPqvRbXeNYSq1gujKz"
   recaptcha_private_key passwords["recaptcha"]


Tigerfell:
Quoting https://github.com/openstreetmap/chef/pull/232#issuecomment-486123587:
> Wait... Will this not require a rebuild?!?

Well, I tried out renaming in a test wiki based on https://www.mediawiki.org/wiki/MediaWiki-Vagrant after creating some pages. It did not break, just the links needed an update or a redirection which I can do manually.

Firefishy:
What about the other wikis we run using this codebase? eg: wiki.osmfoundation.org

Tigerfell:
Is the wiki cookbook not used for the main wiki only? If I understand it correctly, we have
* mediawiki cookbook for all MediaWiki instances
* wiki cookbook for main wiki
* foundation cookbook for foundation wiki 

and so on.

Firefishy:
Merged and live.

Tigerfell:
Unfortunately, this did not work out. We still have the same name. In the end, the PHP config file should include the line ```$wgMetaNamespace = 'Wiki';```. 
Does Chef update the PHP config files automatically or do you need to trigger something? In Vagrant I could change the line in PHP and I hoped that Chef would do that as well.

tomhughes:
It was never actually live - it looks like the merge was only ever pushed to the github version of the repo and not to the production one.

Firefishy:
Sh*t sorry, that was my fault.

Firefishy:
I can confirm `$wgMetaNamespace = 'Wiki';` is now live on wiki.osm.org

PR #241, opened 2019-06-15
add MultiMaps extension to wikis

Tigerfell:
Closes openstreetmap/operations#252 by adding MultiMaps to all wikis. The following steps are needed to finally replace unmaintained SlippyMap extension

- [ ] @gravitystorm obtain an API key from thunderforest.com for displaying transport tiles
- [ ] merge this PR
- [ ] replace all calls to SlippyMap with MultiMaps (I am not sure if the extension was used on any of the private wikis at all, but it was installed) - I can do that for the main wiki using a bot, but I will have to follow Automated edits code of conduct meaning it will take its time. You might want to consider if map extensions are needed in the private wikis, in case they are not used. 
- [ ] remove SlippyMap extension from the configuration

### DIFF
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -430,6 +430,13 @@ action :create do
     update_site false
   end
 
+  mediawiki_extension "MultiMaps" do
+    site new_resource.site
+    template "mw-ext-MultiMaps.inc.php.erb"
+    reference "master"
+    update_site false
+  end
+
   mediawiki_extension "Mantle" do
     site new_resource.site
     update_site false
--- a/cookbooks/mediawiki/templates/default/mw-ext-MultiMaps.inc.php.erb
+++ b/cookbooks/mediawiki/templates/default/mw-ext-MultiMaps.inc.php.erb
@@ -1,0 +1,17 @@
+<?php
+# DO NOT EDIT - This file is being maintained by Chef
+
+require_once "$IP/extensions/MultiMaps/MultiMaps.php";
+
+# Array of String. Array containing all the mapping services that will be made available to the user.
+# First value - default service, which will be used if the service is not in the parameters
+# Values may be a valid name of class based on class BaseMapService or some string and an array if they 
+# denote different tiles within a BaseMapService
+$egMultiMaps_MapServices = [
+  'Leaflet',
+  'transport' => [
+    'service' => 'Leaflet',
+    'attribution' => '&copy; <a href="https://osm.org/copyright">OpenStreetMap contributors</a>. Tiles courtesy of <a href="https://www.thunderforest.com/" target="_blank">Andy Allan</a>',
+    'source' => 'https://tile.thunderforest.com/transport/{z}/{x}/{y}.png',
+  ],
+];


tomhughes:
Have the issues discussed in the previous ticket been resolved?

tomhughes:
So is this creating two services or one? Is `Leaflet` a service? Presumably that is the default OSM tiles?

Where exactly does the API key need to go - is there a specific setting for it or do I just add it into the URL for the transport layer?

Tigerfell:
> Have the issues discussed in the previous ticket been resolved?

In case of openstreetmap/operations#252: yes, in case of openstreetmap/operations#249: The problem was the use of composer, which is not used here. 

> So is this creating two services or one? Is `Leaflet` a service? Presumably that is the default OSM tiles?

This uses two tile layers. Historically, you could choose to have something like ```google``` and ```leaflet``` which would offer Google Maps and OSM standard tiles served in their own framework each (leaflet in case of OSM). The wiki users then have the option to write ```service=leaflet``` to display OSM. I expanded the extension to allow different tiles (like Thunderforest's transport tiles). This is defined in the array. The transport tile layer uses Leaflet, but with a different tile address and attribution. 
> 
> Where exactly does the API key need to go - is there a specific setting for it or do I just add it into the URL for the transport layer?

There is no setting just add it into the URL, please.

pnorman:
> I am not sure if the extension was used on any of the private wikis at all, but it was installed

I've not seen usage on the private wikis I'm on.

Tigerfell:
Thank you for adding the extension. Now we have version 0.7.2, but we need 0.7.3 (newest version) to include my addition to the code. I tried to implement this writing `version 'master'` in cookbooks/wiki/recipes/default.rb. Otherwise, Thunderforest's tiles will not be available (0.7.2 can not handle multiple tile addresses). 
List of commits to this extension and its branches: https://gerrit.wikimedia.org/g/mediawiki/extensions/MultiMaps

tomhughes:
We always use the branch that corresponds to the mediawiki version we are running, both to ensure compatibility and to ensure we get updates as they are released.

tomhughes:
If you're sure that master will work with 1.31then we could consider it, or even update to 1.33 if necessary, but I'm reluctant because we will need to remember to change it back when the necessary edits make it to a release branch and I doubt we will...

tomhughes:
Actually 1.33 isn't released yet but 1.32 would be posssible.

Tigerfell:
Can we please give the master branch a try? The extension's maintainer tried to backport the changes in Wikimedia Gerrit, but unfortunately there is no one with the capacity to merge the change for two weeks now (https://gerrit.wikimedia.org/r/#/c/mediawiki/extensions/MultiMaps/+/519538/; the code review is negative because the maintainer created an indentation error) and the backport would have copied the code from master to the other branches only anyway. 
I would add a note to the calendar for June 2020 (assuming that we continue with 1.31).

tomhughes:
I'm really reluctant because if we do that (a) we'll be tracking all upstream changes as they happen so things could break at any time and (b) we'll have to remember to change it back once there is a proper release.

Tigerfell:
@pastakhov Can you help us here?

PR #246, opened 2019-08-21
Added new Air3 version

natrius:


natrius:
@tomhughes 
Sorry for bothering, but when do the chef-scripts get executed? Every 24 hours or something like this?
If it should happened already, it did not work - maybe is there a "overwrite" or something like this needed?

tomhughes:
It runs every thirty minutes but, as you say, it needed to allow overwrites. Fixed in 0d8030b3a7ec9eaa6bc96c7dd39d835a00a7ace0 and run manually to fix things.

PR #248, opened 2019-09-03
Resolve multiple Cookstyle warnings

tas50:
Chef specific rules are being added to Cookstyle to replace Foodcritic as the single Ruby + Chef linter in the Chef ecosystem. I ran the current master of Cookstyle over this repo with autocorrect enabled.

What changed
- Removed the unused long_description metadata from cookbook metadata.rb
- Converted all node attributes from symbols to strings
- Converted file modes to strings

Signed-off-by: Tim Smith <tsmith@chef.io>

### DIFF
--- a/cookbooks/accounts/metadata.rb
+++ b/cookbooks/accounts/metadata.rb
@@ -3,6 +3,5 @@ maintainer        "OpenStreetMap Administrators"
 maintainer_email  "admins@openstreetmap.org"
 license           "Apache-2.0"
 description       "Accounts management"
-long_description  IO.read(File.join(File.dirname(__FILE__), "README.md"))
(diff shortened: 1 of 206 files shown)


tomhughes:
We need to actually switch to [cookstyle](https://github.com/chef/cookstyle) before this makes much sense, or the changes are just going to bitrot.

As I understand it the idea is that cookstyle replaces both rubocop and foodcritic? Is it actually ready to fully make that switch now or would you recommended leaving foodcritic enabled as well for now?

Is there any documentation anywhere that discusses the rationale behind the recommendations that lead to the changes being suggested here?

tas50:
Closing this out in favor of a full cookstyle PR

PR #250, opened 2019-09-06
Extend removal of the interface-admin user group to the main wiki

Tigerfell:
#249 copied the permissions of the interface-admin user group to the administrators and subsequently removed the IA, but due to a lack of formatting, we missed that it was paced in a conditional block, so it was applied to private wikis only. I indented all those blocks and moved the addition out of the if-block. @nyurik sorry, I did not see that in the first place.

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -113,9 +113,9 @@ $wgSVGMaxSize = 2000;
 
 # InstantCommons allows wiki to use images from https://commons.wikimedia.org
 <% if @mediawiki[:commons] -%>
-$wgUseInstantCommons  = true;
+	$wgUseInstantCommons  = true;
 <% else -%>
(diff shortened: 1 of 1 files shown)


tomhughes:
I understand the logic of indenting but it does mean the generated file is a complete mess which is why we generally don't do that...

Tigerfell:
How about a single space indent then? I just followed the indentation of https://github.com/openstreetmap/chef/blob/03b5f3ea1de0f4490857a894b407bf4f314a5d2c/cookbooks/mediawiki/templates/default/LocalSettings.php.erb#L98-L100

tomhughes:
Sorry, but I'm afraid I took @nyurik's version because I stared at the diff for this one for five minutes and was struggling to understand what i was doing... I think it is right but it was much easier to tell with the other one.

Tigerfell:
It does exactly the same thing as #251, but also adds indentation. Looks like the diff can not display the addition of a single character at the start of the line properly.

PR #253, opened 2019-09-07
Switch from RuboCop/Foodcritic to Cookstyle

tas50:
Swap rubocop, rubocop-performance, and foodcritic for cookstyle, which has a ruleset tailored for Chef cookbooks as well as 76 Chef specific rules. In doing this I wiped the .rubocop.yml file to just use the Chef standard rules. This standardizes the OSM repo on the same linting guidelines used by the Chef community and also speeds up the overall CI run since it no longer requires the large number of compiled gems.

I removed Foodcritic as it's about 90% redundant at this point, and it has a very high false positive rate. Cookstyle has 76 Chef rules at this point and Foodcritic has a bit over 100. In the next few weeks we'll probably have about the same number of checks in Cookstyle, but without the false positives.

The [Cookstyle readme](https://github.com/chef/cookstyle/blob/master/README.md) outlines the rational for creating Cookstyle instead of using Rubocop. There's also information on each of the Chef specific cops in the [Cops docs](https://github.com/chef/cookstyle/blob/master/docs/cops.md).

Let me know if there's anything else I can do here. I'm around.

Signed-off-by: Tim Smith <tsmith@chef.io>


tomhughes:
Thanks.

I'll have a look in detail later but I'm not sure I like removing the existing rubocop.yml as most things in there represent deliberate policy choices in our part and align with our other ruby based repositories even if they're not necessarily considered best practice these days.

I'm thinking in particular of things like quoting and hash style - at the very least if we decide to change those we probably want to do it across all our repositories and as a separate action to switching to chefstyle here.

tas50:
Take a look now. I force pushed to avoid a large repo size.

tomhughes:
Thanks a lot for the PRs and especially for pointing us at cookstyle which I wasn't previously aware of.

Rather than take everything as one big change however I've decided to take a more gradual approach to switching so that I can properly evaluate each change and understand what is being changed and why.

I have just pushed a basic commit to switch us to cookstyle while preserving out existing ruby formatting rules and committing any ChefStyle rules which are being violated to the todo file for gradual consideration, followed by commits to fix some of the basic and fairly risk free style fixes.

There's about five things left that I need to consider - two of them (string keys and file modes) are relatively straightforward and risk free if large in scope and the others will involve some more investigation of what change is being proposed and whether it is likely to break anything.

I'd still be interested in knowing if there is any documentation anywhere of the rationale behind the rules (like string keys and file modes) which aren't particularly about correctness - or are they?

PR #257, opened 2019-10-27
Create "Proposal" namespace for main wiki

Tigerfell:
This PR configures a new namespace "Proposal" for the main. It will have a slightly lower search rank as the tag/key pages. Additionally, it moves the namespace config to a new file in the wiki cookbook. This will remove the language namespaces for all other wikis.
Before deploying please check

- [ ] none of the other wikis should have any pages in the language namespaces like "FR" or "FR talk". You can use [Special:PrefixIndex](https://wiki.openstreetmap.org/wiki/Special:PrefixIndex).
- [ ] that the job queues of the wikis are empty as explained in https://www.mediawiki.org/wiki/Manual:Using_custom_namespaces
- [ ] temporarily remove the two pages starting with the same prefix "Proposal:" in the main wiki. You can restore them after the config change.

I created four commits, so you can squash them together as you like. 

Related discussion https://wiki.openstreetmap.org/wiki/Talk:Wiki/Archive_6#Proposal:_new_namespace_for_tagging_proposals and again starting with https://lists.openstreetmap.org/pipermail/talk/2019-April/082453.html. I asked in mid-August if someone would object when I would open this PR and nothing happened.


Tigerfell:
@Firefishy ?

Firefishy:
Sorry, I think I missed this earlier.
I am unsure how `cookbooks/wiki/templates/default/LocalSettings.php.erb` would be loaded. I don't think chef would load that file currently.

Needs some work, but looks good.

Tigerfell:
> I am unsure how `cookbooks/wiki/templates/default/LocalSettings.php.erb` would be loaded. I don't think chef would load that file currently.

Yes, that is correct. The idea is to have an additional `LocalSettings.php` for the main wiki which adds some specific configuration. You would need to adapt your internal chef deployment accordingly.

Tigerfell:
@Firefishy Let's try to figure this out. I tried to implement a [suggestion](https://github.com/openstreetmap/operations/issues/302#issuecomment-495954615) by @tomhughes from a different issue: 
> Well just put it in the `LocalSettings` template, or as a new dropin file in `LocalSettings.d`.

Now I see the file is supposed to be named differently, but I do not know which is the correct folder in your setup. My test wiki has a folder `settings.d` and it executes all PHP files inside.

cookbooks/wiki/templates/default/LocalSettings.php.erb
```diff
@@ -0,0 +1,90 @@
+<?php
```
tomhughes: Either this template is needed, in which some recipe needs to instantiate it, or it's not in which case it should be removed.

HolgerJeromin:
> (technical reason: I could not change the name of the branch in GitHub).

JFYI: You could have set the **upstream** branch for your local new branch (`pr257`) to `patch-2` and force push to this old one. This would update/recycle this PR.

Tigerfell:
Thanks, @HolgerJeromin !

PR #259, opened 2020-01-15
Fixed small bugs after trying wiki test-kitchen

migurski:
- Updated Ubuntu from 16.04 to 18.04 based on [current running hardware](https://hardware.openstreetmap.org)
- ~~Added Mediwaiki Apt repository to support Parsoid install~~ Added `apt` as `mediawiki` dependency to complete installation
- ~~Added missing Wiki user account~~ Added `accounts` as `mediawiki` dependency to complete installation
- Added dummy passwords in test data bag

Test plan:
---
1. Add test kitchen configuration like:

          - name: migurski_testing
            run_list:
              - recipe[munin]
              - recipe[wiki]
              - role[wiki]

2. Run `kitchen converge migurski-testing-ubuntu-1804`

Result
---
Still seeing some late-stage failures detailed in this log, but gets further than previously: https://gist.github.com/migurski/e08c0207d18462258498b198f1ae3607

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -323,3 +323,10 @@ suites:
   - name: wordpress
     run_list:
       - recipe[wordpress::default]
+  - name: wiki
+    run_list:
+      - recipe[apt::default]
+      - recipe[accounts::default]
+      - recipe[munin::default]
+      - recipe[wiki]
+      - role[wiki]
--- a/.github/workflows/test-kitchen.yml
+++ b/.github/workflows/test-kitchen.yml
@@ -92,6 +92,7 @@ jobs:
           - web-frontend
           - web-rails
           - wordpress
+          - wiki
         os:
           - ubuntu-1804
       fail-fast: false
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -397,6 +397,8 @@ action :create do
   mediawiki_extension "CheckUser" do
     site new_resource.site
     template "mw-ext-CheckUser.inc.php.erb"
+    repository "git://github.com/migurski/extension-checkuser.git"
+    reference "0fe3015e57bb6ee771ecaebf3b2d79a251bd4e3a"
     update_site false
   end
 
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -398,7 +398,7 @@ action :create do
     site new_resource.site
     template "mw-ext-CheckUser.inc.php.erb"
     repository "git://github.com/migurski/extension-checkuser.git"
-    reference "0fe3015e57bb6ee771ecaebf3b2d79a251bd4e3a"
+    reference "0e3bf4aa3d05b60240b0abc1a00d87c9fc4c5cf7"
     update_site false
   end
 
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -397,8 +397,6 @@ action :create do
   mediawiki_extension "CheckUser" do
     site new_resource.site
     template "mw-ext-CheckUser.inc.php.erb"
-    repository "git://github.com/migurski/extension-checkuser.git"
-    reference "0e3bf4aa3d05b60240b0abc1a00d87c9fc4c5cf7"
     update_site false
   end
 
--- a/test/data_bags/accounts/wiki.json
+++ b/test/data_bags/accounts/wiki.json
@@ -1,0 +1,7 @@
+{
+  "id": "wiki",
+  "uid": "555",
+  "comment": "wiki.openstreetmap.org",
+  "home": "/opt/wiki",
+  "manage_home": false
+}
--- a/test/data_bags/wiki/passwords.json
+++ b/test/data_bags/wiki/passwords.json
@@ -1,0 +1,6 @@
+{
+  "database": "pw1",
+  "admin": "pw2",
+  "recaptcha": "pw3",
+  "thunderforest": "pw4"
+}


cookbooks/mediawiki/recipes/default.rb
```diff
@@ -22,6 +22,18 @@
...
 
+# Mediawiki Parsoid package is available from wikimedia.org repositories,
+# more information at https://www.mediawiki.org/wiki/Parsoid/Setup#Ubuntu_/_Debian
+apt_repository 'mediawiki' do
+  uri           'https://releases.wikimedia.org/debian'
+  distribution  'jessie-mediawiki'
+  components    ['main']
+  key           'AF380A3036A03444'
+  keyserver     'keyserver.ubuntu.com'
+end
```
tomhughes: This is not how we configure APT in our environment - the repositories are configured via the `apt` cookbook so it's not acceptable to try and do it here as well.

The problem is you're trying to use one cookbook on it's own and that just doesn't work for any non-trivial cookbook in our setup.

cookbooks/mediawiki/recipes/default.rb
```diff
@@ -22,6 +22,18 @@
...
+
+user 'wiki'
```
tomhughes: Again, users are configured by the `account` cookbook.

cookbooks/mediawiki/recipes/default.rb
gravitystorm: > The problem is you're trying to use one cookbook on it's own and that just doesn't work for any non-trivial cookbook in our setup.

I think that's a cop-out. All the cookbooks should work, trivial or not. If they have hidden dependencies on other cookbooks, then those dependencies should be made explicit.

>  so it's not acceptable to try and do it here as well.

I'd prefer to see a suggestion of how you would prefer to see the problem solved, rather than just saying something is "not acceptable". For example, if running the apt recipe is needed first, then you could suggest adding an 'include_recipe' statement.

tomhughes: Well he hasn't really explained what he was doing - there is no test currently declared for this cookbook so presumably he was specifying something on the command line somehow?

Obviously the depedncency should be noted in `metadata.rb` though that isn't actually enough to make it run, and I'm not sure it's safe or necessarily correct or appropriate to just `include_recipe` it - our whole system is built around the roles and assumes that the base roles and the cookbooks they require will be run everywhere. Fixing (or changing if you don't accept that the current state is broken as such) that is a massive job.

migurski: My test plan in the PR upthread explains that I was fixing some issues that I had run into when trying to run the wiki cookbook under test-kitchen. Thanks for your suggestions, I see now that the missing Apt repository is specified in a different cookbook that wasn't declared as a dependency in this one. I also noticed that it was necessary to have `recipe[munin]` in my test configuration; I’ll modify this PR to account for your suggestions.

tomhughes: My point was that `.kitchen.yml` doesn't contain any tests for the wiki cookbook, or are you tryimg to add one?

migurski:
I made some edits in 45423046 that specify the `apt` and `account` cookbooks as dependencies of `mediawiki` based on requested changes. It seems like `munin` might also be a dependent cookbook in this case; I’ll need some more time to re-run from scratch and test that change before making further edits to this PR.

cookbooks/mediawiki/recipes/default.rb
migurski: I’ll go ahead and add one — I was using the one described in my test plan above, but it sounds from your suggestions like it would be appropriate to add one like:

      - name: wiki
        run_list:
          - recipe[wiki]
          - role[wiki]

tomhughes: I wouldn't expect to name both the recipe and the role like that - does the role not pull in the recipe anyway? You will also need to pull in the dependent recipes most likely - look at the otrs test for example.

I suspect the wiki one probably has quite complex dependencies which is likely why nobody has been in a hurry to tackle it.

migurski:
In 638131da2df615c77c5e6f00d79a58d82426563e I completed the addition of the "wiki" user account in the `accounts` cookbook and updated the correct signing key for the Mediawiki apt repository. Munin seemed to be a pre-requisite as well, and there’s now a configuration in test-kitchen for the wiki so this works:

    kitchen converge wiki-ubuntu-1804

The Mediawiki cookbook still seems buggy to me. On a fresh run, there’s a SQL error which will disappear on a second run: https://gist.github.com/migurski/e2096952dbff2779223751d9ab5fa7d5 — this non-idempotent behavior makes me very uncomfortable.

tomhughes:
Pretty sure it worked fine last time I deployed a new wiki but mediawiki could easily have changed/broken something in between. They do like making life hard for us so it's all a bit hit and miss how well things work I'm afraid.

cookbooks/accounts/recipes/default.rb
```diff
@@ -107,3 +107,6 @@
   action :modify
   members administrators.sort
 end
+
+# Required account for mediawiki cookbook
+user "wiki"
```
tomhughes: This is not how you add an account! Please look at how the accounts cookbook works to understand what you need to do... I would expect it will just happen though if you have the wiki role loaded and run the `accounts::default` recipe.

cookbooks/apt/recipes/default.rb
```diff
@@ -158,7 +158,7 @@
   uri "https://releases.wikimedia.org/debian"
   distribution "jessie-mediawiki"
   components ["main"]
-  key "90E9F83F22250DD7"
+  key "AF380A3036A03444"
```
tomhughes: If the key has changed, which I doubt as we would be getting errors, then this should be fixed in a separate pull really.

cookbooks/mediawiki/recipes/default.rb
```diff
@@ -17,6 +17,9 @@
 # limitations under the License.
 #
 
+include_recipe "apt"
+include_recipe "accounts"
+include_recipe "munin"
```
tomhughes: As I said earlier, this is probably not the best way to do it - there are certain recipes are automatically set on all hosts so we don't include them in every cookbook like this. Please, please, look at how the other tests are doing this.

cookbooks/accounts/recipes/default.rb
migurski: I’ll update, thanks. It didn’t just work, unfortunately. The ordering of the role/recipe run list is significant and I’ll reverse them and see if it helps and makes this line unnecessary.

cookbooks/apt/recipes/default.rb
migurski: According to https://www.mediawiki.org/wiki/Parsoid/Setup#Ubuntu_/_Debian, "AF380A3036A03444" is the correct key. "90E9F83F22250DD7" did not work for me when testing: Parsoid came up as an unavailable package.

cookbooks/mediawiki/recipes/default.rb
migurski: Hm, this looked to be a pattern in use in other tests. I see a mix of approaches: some tests have additional recipes in their run list while others include recipes as I’m doing here. Can you point to a specific pattern I should follow?

tomhughes: Look at the `otrs` test for example - it runs extra recipes directly in the `.kitchen.yml` definition.

As a general rule if you're having to alter the cookbooks to test them then there's probably something wrong somewhere...

cookbooks/apt/recipes/default.rb
tomhughes: You're quite right - they updated it recently it seems. I though that `apt-get update` exited non-zero when that happens, which makes it quite noticeable, but apparently it just skips updating that repo :-(

I've pushing a fix for that now.

cookbooks/accounts/recipes/default.rb
migurski: Following up, this really didn’t help and [immediately brought up an elasticsearch error](https://gist.github.com/migurski/fad477dbf80bc78bef1bfeba80397bdb):

         run_list:
           - role[wiki]
           - recipe[wiki]

cookbooks/mediawiki/recipes/default.rb
migurski: Thanks, I’ll dig into `otrs`!

migurski:
I moved to a style that I think matches what’s in `otrs`: recipes for `apt`, `accounts`, and `munin` all added to the test configuration and removed from dependencies within the `wiki` and `mediawiki` cookbooks. The new configuration is still not picking up the `wiki` user by default, and dies here regardless of the role/recipe order in the run list: 

    AL: Chef::Exceptions::UserIDNotFound: mediawiki_site[wiki.openstreetmap.org] (wiki::default line 30) had an error: Chef::Exceptions::UserIDNotFound: directory[/srv/wiki.openstreetmap.org] (/tmp/kitchen/cache/cookbooks/mediawiki/resources/site.rb line 94) had an error: Chef::Exceptions::UserIDNotFound: cannot determine user id for 'wiki', does the user exist on this system?

I’ve also tried moving `role[wiki]` to the top of the wiki configuration run list but this surfaces a problem in the Elasticsearch default attributes `undefined method '|' for {}:Chef::Node::VividMash` at:

https://github.com/openstreetmap/chef/blob/45dde9418dd342bb1a632d82559201c0b3becf1a/cookbooks/elasticsearch/attributes/default.rb#L7

tomhughes:
Have you added `test/data_bags/accounts/wiki.json` to define the wiki account?

migurski:
Good hint. I did that, and then found an early fatal error due to a missing `node[:etc]` value which I fixed in the default attributes in e87c666c32ede8d3f5814a8a109c495709548de4. Unsure if this is the right way to go about supplying values, or if the cookbook should be more resilient to missing values?

gravitystorm:
> Unsure if this is the right way to go about supplying values, or if the cookbook should be more resilient to missing values?

As I commented earlier, the right way to do things is to make each cookbook or role converge on its own. Putting dependent recipes in the test-kitchen config is just a workaround for these missing dependencies.

I don't understand what advantage there is to having all these missing dependencies, but from the comments above, @tomhughes seems quite keen on keeping it that way. :man_shrugging: 

> As a general rule if you're having to alter the cookbooks to test them then there's probably something wrong somewhere...

Yes, the problem is the cookbooks need to be fixed in order to include their implicit dependencies! Working around that (via the test-kitchen config) is a workaround - and not something we should perpetuate when a volunteer is volunteering to fix it properly by using `include_recipe`.

Still, it's @tomhughes who gets the final say on this topic, so if we can't persuade him otherwise, I guess we keep layering on the workarounds.

tomhughes:
So first I would need to study this in more detail - to be honest I'm not entirely sure exactly what `include_recipe` does as such and what the implications of it potentially being pulled in more than once are. I assume it's much the same as listing the recipe in the role except that a topological sort is done to order the overall set of recipes correctly which should be fine.

The larger problem is that many of our recipes make no sense, or won't work at all, without some sort of configuration - so there is no point for example pulling in the recipe to configure the network without the host role that contains the attributes describing the network configuration for the machine. That's an extreme example perhaps as not many things will need to pull that in but the general scope of this problem is mind boggling.

I realise that our chef configuration is not at all modern, and may not even be done well on it's own terms, but fixing that is a separate issue to writing tests - we shouldn't try and redesign our chef setup as a side effect of adding a test, it should be a separate task.

The trickiest issues do generally revolve around shared resources like users and apt repositories which is why we configure them in slightly odd ways and why you can't just use a resource in the way @migurski was initially trying to do. That said for relatively simple cases like here just including the recipes is probably enough so long as you at least have the `wiki` role which turns on the right accounts and repositories.

As an aside I would just point out that the `wiki` cookbook is probably not the best place to start - that is specific to configuring `wiki.openstreetmap.org` and it would probably be better to start with the `mediawiki` cookbook which handles generic installation of a mediawiki instance - the `wiki` cookbook then builds on that to create one specific wiki.

Generally speaking I would love to have tests, although I remain unconvinced that the kind of granularity @gravitystorm is practical and/or useful - mostly my concern is not "will cookbook X run in some default configuration" but "will the total set of things deployed on machine X work" which is a rather different question. That said individual cookbook tests is a good start I just worry about having to duplicate things to make it work and that actually increasing the risk in real deployments from having conflicting information.

Really the main reason I haven't pursued tests more though is that running them turns out to be such a nightmare, and so slow, that I find them impractical to use in real life. That's why we have so few and why I stopped after writing a few of the simpler ones.

gravitystorm:
> As an aside I would just point out that the `wiki` cookbook is probably not the best place to start

The reason that I've been suggesting the wiki cookbook is that it's the system that has the most outside interest, i.e. from non-sysadmins, over the last 3 years. We have people who want to change the configuration of wiki.openstreetmap.org, and it would be easier for them if they could converge the recipes locally, and easier for admins if it was easy to check PRs locally before deploying them. But without the ability to test convergence using test-kitchen, outside contributors are coding blind, PRs are hard to review, and we only find out if the changes actually work when we run them on the live servers.

Sure, if test coverage was my goal, I'd suggest starting elsewhere! But my goal is to get other people involved in this repo, so targetting the cookbook with the most outside interest seems like the right approach, since it's the most likely one to get external PRs.

> mostly my concern is not "will cookbook X run in some default configuration" but "will the total set of things deployed on machine X work" which is a rather different question.

That's totally understandable. But for other contributors wanting to try writing some code, it's probably better to have smaller testable options (like testing individual cookbooks), rather than having to run an all-up integration test each time they change a line of code.

> Really the main reason I haven't pursued tests more though is that running them turns out to be such a nightmare, and so slow, that I find them impractical to use in real life.

Sure, it's definitely an overhead if you're happy to commit to the repo and run the changes on the live servers to check that they work. And without splitting the repo up into smaller parts, it's unrealistic to run any CI service (imagine having it converge every cookbook when a PR only affects one of them!). And we haven't even touched on using test-kitchen to actually do some tests, since we're so far only checking if the code converges as a binary yes/no, and not doing anything like using chefspec or a kitchen verifier to make sure e.g. services are running or throwing configuration errors.

But having the recipes and roles converging locally - as a bare minimum step - is still useful and helps other people get involved, and that's what I think we should work together on.

tomhughes:
> The reason that I've been suggesting the wiki cookbook is that it's the system that has the most outside interest, i.e. from non-sysadmins, over the last 3 years. We have people who want to change the configuration of wiki.openstreetmap.org, and it would be easier for them if they could converge the recipes locally, and easier for admins if it was easy to check PRs locally before deploying them. But without the ability to test convergence using test-kitchen, outside contributors are coding blind, PRs are hard to review, and we only find out if the changes actually work when we run them on the live servers.

I was really just trying to say that the `wiki` cookbook probably isn't what people think it is - most of the configuration of the wiki is actually in the `mediawiki` cookbook.

The biggest problem with the wiki anyway is not having anybody that is confident enough in mediawiki to be able to review the PRs and requests.

> Sure, it's definitely an overhead if you're happy to commit to the repo and run the changes on the live servers to check that they work. And without splitting the repo up into smaller parts, it's unrealistic to run any CI service (imagine having it converge every cookbook when a PR only affects one of them!). And we haven't even touched on using test-kitchen to actually do some tests, since we're so far only checking if the code converges as a binary yes/no, and not doing anything like using chefspec or a kitchen verifier to make sure e.g. services are running or throwing configuration errors.

Actually some of them do test that the right services are running afterwards!

My real point though is that whenever I try and run them I spend the first three hours trying to convince vagrant to work and then every test run takes about ten minutes because it has to spin up a VM which it then probably forgets to shutdown again so I find them cluttering up my system weeks later :-(

I wish I knew a good solution to this!

gravitystorm:
> Actually some of them do test that the right services are running afterwards!

Awesome :-)

> then probably forgets to shutdown again

Yeah, running 'converge' leaves the VM available for inspection, you need to run 'destroy' to get rid of it. That's much less frustrating than running 'test', which destroys the vm automatically at the end of the run - that makes development super slow! 

> I wish I knew a good solution to this!

I use container-based drivers (kitchen-docker, see also kitchen-dokken) whenever possible, since they are faster to spin up than a VM, but there's some networking edge cases that they can't handle. So they are more useful for individual cookbook testing and less useful for all-up role tests. We can mix and match them in the kitchen config - individual suites can have different drivers.

mmd-osm:
> I spend the first three hours trying to convince vagrant to work and then every test run takes about ten minutes because it has to spin up a VM which it then probably forgets to shutdown again so I find them cluttering up my system weeks later :-(

You can try a Docker based approach I described in more detail here: https://github.com/openstreetmap/chef/pull/74#issuecomment-421725316

It's much more lightweight than the virtualbox.

tomhughes:
Well frankly I hate docker as well, though in theory podman should be able to stand in and avoid many of it's problems. I am working on proving that theory - so far there are a few edge cases with the compatibility.

I think originally we came to the conclusion that containers wouldn't work because they didn't have an init so you couldn't start services or anything. I know a container can gave an init but is kitchen actually able to use it in that way?

Finally I am somewhat concerned that all the URLs to kitchen.ci on the kitchen github are dead - the domain no longer exists. Is kitchen dead technology?

migurski:
The kitchen.ci DNS failure also gives me pause. Maybe there will be an update on https://ourincrediblejourney.tumblr.com/ … 

Reading back through this thread it sounds like my PR is still an incremental step forward even though it doesn’t quite address all problems. I am happy to continue poking at this if it is useful. I wonder if we can merge this set of changes before starting the next set? I’m interested to hear a resolution on the `include_recipe` question for example. I definitely agree that being able to run a selected piece of OSM’s infra locally in order to make changes is vital for ensuring quality community suggestions.

As far as testing goes, I am tinkering with getting these tests running on a clean Ubuntu system. Right now I’m struggling to get past an early error:

    Vagrant was unable to communicate with the guest machine within
    the configured ("config.vm.boot_timeout" value) time period.

This could be configured to run in parallel on a temporary EC2 server where it might be integrated into the CI for this repository?

migurski:
Poking this thread another time: I think I’ve made the requested changes, and this PR incrementally improves the repo even if the `wiki` cookbook does not yet fully complete.

tomhughes:
I'm still working on it but I need to get my test kitchen environment up and working again before I can look at this in detail.

mmd-osm:
For the fun of it, I was following the steps in https://www.openstreetmap.org/user/migurski/diary/391942 on Ubuntu 18.04 / virtualbox. Running the script took about 25 minutes to run and failed with the same message as mentioned for Docker in the blog post:

```
       Creating cu_log table ...done.
       ...index cuc_ip_hex_time already set on cu_changes table.
       ...index cuc_user_ip_time already set on cu_changes table.
       ...have cuc_private field in cu_changes table.
       ...site_stats is populated...done.
       ...Update 'populate rev_len and ar_len' already logged as completed.
       ...Update 'populate rev_sha1' already logged as completed.
       ...img_sha1 column of image table already populated.
       ...protocol-relative URLs in externallinks table already fixed.
       ...fa_sha1 column of filearchive table already populated.
       ...*_from_namespace column of backlink tables already populated.
       ...Update 'FixDefaultJsonContentPages' already logged as completed.
       ...Update 'cleanup empty categories' already logged as completed.
       ...RFC and PMID already added to interwiki database table.
       ...Update 'populate pp_sortkey' already logged as completed.
       ...Update 'populate ip_changes' already logged as completed.
       ...externallinks table indexes up to date
       Starting poulation of cu_changes with recentchanges rc_id from 1 to 100
       ...migrating rc_id from 1 to 100
       STDERR: [ccaedd880af180fc311d3c19] [no req]   Wikimedia\Rdbms\DBQueryError from line 1587 of /srv/wiki.openstreetmap.org/w/includes/libs/rdbms/database/Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading? 
       Query: INSERT  INTO `cu_changes` (cuc_timestamp,cuc_user,cuc_user_text,cuc_namespace,cuc_title,cuc_comment,cuc_minor,cuc_page_id,cuc_this_oldid,cuc_last_oldid,cuc_type,cuc_ip,cuc_ip_hex) VALUES ('20200119093941',NULL,'MediaWiki default','0','Main_Page','','0','1','1','0','1','127.0.0.1','7F000001')
       Function: PopulateCheckUserTable::doDBUpdates
       Error: 1048 Column 'cuc_user' cannot be null (localhost)
       
       Backtrace:
       #0 /srv/wiki.openstreetmap.org/w/includes/libs/rdbms/database/Database.php(1556): Wikimedia\Rdbms\Database->getQueryExceptionAndLog(string, integer, string, string)
       #1 /srv/wiki.openstreetmap.org/w/includes/libs/rdbms/database/Database.php(1274): Wikimedia\Rdbms\Database->reportQueryError(string, integer, string, string, boolean)
       #2 /srv/wiki.openstreetmap.org/w/includes/libs/rdbms/database/Database.php(2149): Wikimedia\Rdbms\Database->query(string, string)
       #3 /srv/wiki.openstreetmap.org/w/extensions/CheckUser/maintenance/populateCheckUserTable.php(98): Wikimedia\Rdbms\Database->insert(string, array, string)
       #4 /srv/wiki.openstreetmap.org/w/maintenance/Maintenance.php(1719): PopulateCheckUserTable->doDBUpdates()
       #5 /srv/wiki.openstreetmap.org/w/maintenance/update.php(215): LoggedUpdateMaintenance->execute()
       #6 /srv/wiki.openstreetmap.org/w/maintenance/doMaintenance.php(96): UpdateMediaWiki->execute()
       #7 /srv/wiki.openstreetmap.org/w/maintenance/update.php(275): require_once(string)
       #8 {main}
       ---- End output of php maintenance/update.php --quick ----
       Ran php maintenance/update.php --quick returned 255
       
-----> Destroying <wiki-ubuntu-1804>...
       ==> default: Forcing shutdown of VM...
       ==> default: Destroying VM and associated drives...
       Vagrant instance <wiki-ubuntu-1804> destroyed.
       Finished destroying <wiki-ubuntu-1804> (0m4.46s).
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Converge failed on instance <wiki-ubuntu-1804>.  Please see .kitchen/logs/wiki-ubuntu-1804.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

mmd-osm:
By the way, a good way to speed things up a bit is to add extra CPUs and memory:

```
platforms:
  # Ubuntu 18.04 is what’s used on the live OSM servers,
  # according to https://hardware.openstreetmap.org/
  - name: ubuntu-18.04
    driver:
      customize:
        cpus: 4
        memory: 4096
```

mmd-osm:
So besides the CheckUser extension issue, there's also an issue with generating SSL certificates, which currently fails inside test-kitchen:

```
FATAL: Net::HTTPServerException: mediawiki_site[wiki.openstreetmap.org] (wiki::default line 30) had an error: Net::HTTPServerException: ssl_certificate[wiki.openstreetmap.org] (/tmp/kitchen/cache/cookbooks/mediawiki/resources/site.rb line 489) had an error: Net::HTTPServerException: 404 "Not Found"
```

tomhughes:
Issuing SSL certificates requires co-operation with another machine, but it should just create a self signed one if that's not available.

I repeat though that I am working on this as fast as I can. As well as several evenings last week I have spent this afternoon on it as well. There is really no point spending a lot of time on a new test until the existing ones are working properly again.

mmd-osm:
No worries. I have those tests running in the background, check from time to time how they're doing and post some results here. The idea is to get a rough overview of what kind of issues lie ahead and need some kind of fix eventually.

migurski:
@mmd-osm, can you share any details on how you installed vagrant on Ubuntu to get as far as you did? So far I’ve only been able to get to this stage on Mac OS as the host.

mmd-osm:
I already played with test kitchen and the osm-chef repo last year. It may very well be that I did encounter the same issue as you did and somehow fixed it (quite likely something I found on StackOverflow), but quite honestly I don't recall.

One thing you could look into is RSA/DSS keys for remote log on via SSH. Those tend to cause some issues sometime. Also when you open up virtualbox, there's an option to inspect the log file for the respective VM.

mmd-osm:
Latest update from my side: after temporarily turning off the CheckUser wiki extension and the SSL for Apache, the wiki cookbook actually builds without error:

```
       Recipe: wiki::default
         * mediawiki_site[wiki.openstreetmap.org] action update
           * template[/srv/wiki.openstreetmap.org/w/LocalSettings.php] action create
             - update content in file /srv/wiki.openstreetmap.org/w/LocalSettings.php from 2da3dd to 5024f9
             --- /srv/wiki.openstreetmap.org/w/LocalSettings.php	2020-01-19 18:40:37.410909028 +0000
             +++ /srv/wiki.openstreetmap.org/w/.chef-LocalSettings20200119-2220-1mvw4s7.php	2020-01-19 18:47:34.807503028 +0000
             @@ -331,6 +331,7 @@
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-CirrusSearch.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-DisableAccount.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Skin-Vector.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Scribunto.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Poem.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-AntiSpoof.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-CleanChanges.inc.php');
             @@ -339,6 +340,9 @@
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-ParserFunctions.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-VisualEditor.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-UniversalLanguageSelector.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-MultiMaps.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-CodeEditor.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-CodeMirror.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-LocalisationUpdate.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-DismissableSiteNotice.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Elastica.inc.php');
             @@ -353,15 +357,20 @@
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Skin-MinervaNeue.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-InputBox.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-WikiEditor.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-OsmWikibase.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-MobileFrontend.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Nuke.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Skin-Modern.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-SpamBlacklist.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Skin-CologneBlue.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-osmtaginfo.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Echo.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Babel.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-ConfirmEdit.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-AbuseFilter.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-TimedMediaHandler.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-PdfHandler.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Thanks.inc.php');
       require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Skin-MonoBook.inc.php');
             +require_once('/srv/wiki.openstreetmap.org/w/LocalSettings.d/Ext-Wikibase.inc.php');
           * execute[/srv/wiki.openstreetmap.org/w/maintenance/update.php] action run
             - execute php maintenance/update.php --quick
           * execute[/srv/wiki.openstreetmap.org/w/maintenance/update.php] action run
             - execute php maintenance/update.php --quick
         
       Recipe: elasticsearch::default
         * service[elasticsearch] action restart
           - restart service service[elasticsearch]
       
       Running handlers:
       Running handlers complete
       Chef Infra Client finished, 460/686 resources updated in 26 minutes 04 seconds
       Downloading files from <wiki-ubuntu-1804>
       Finished converging <wiki-ubuntu-1804> (26m52.12s).
-----> Setting up <wiki-ubuntu-1804>...
       Finished setting up <wiki-ubuntu-1804> (0m0.00s).
-----> Verifying <wiki-ubuntu-1804>...
       Preparing files for transfer
       Transferring files to <wiki-ubuntu-1804>
       Finished verifying <wiki-ubuntu-1804> (0m0.00s).
-----> Destroying <wiki-ubuntu-1804>...
       ==> default: Forcing shutdown of VM...
       ==> default: Destroying VM and associated drives...
       Vagrant instance <wiki-ubuntu-1804> destroyed.
       Finished destroying <wiki-ubuntu-1804> (0m4.89s).
       Finished testing <wiki-ubuntu-1804> (27m50.82s).
-----> Test Kitchen is finished. (27m51.01s)
```

Diff:

```
diff --git a/.kitchen.yml b/.kitchen.yml
index 687a142a..d65a0b53 100644
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -13,6 +13,10 @@ platforms:
   # Ubuntu 18.04 is what’s used on the live OSM servers,
   # according to https://hardware.openstreetmap.org/
   - name: ubuntu-18.04
+    driver:
+      customize:
+        cpus: 4
+        memory: 4096
 
 suites:
   - name: accounts
diff --git a/cookbooks/mediawiki/resources/site.rb b/cookbooks/mediawiki/resources/site.rb
index 55228af0..94057f41 100644
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -393,11 +393,11 @@ action :create do
     update_site false
   end
 
-  mediawiki_extension "CheckUser" do
-    site new_resource.site
-    template "mw-ext-CheckUser.inc.php.erb"
-    update_site false
-  end
+#  mediawiki_extension "CheckUser" do
+#    site new_resource.site
+#    template "mw-ext-CheckUser.inc.php.erb"
+#    update_site true
+#  end
 
   mediawiki_extension "DismissableSiteNotice" do
     site new_resource.site
@@ -486,9 +486,9 @@ action :create do
     backup false
   end
 
-  ssl_certificate new_resource.site do
-    domains [new_resource.site] + Array(new_resource.aliases)
-  end
+#  ssl_certificate new_resource.site do
+#    domains [new_resource.site] + Array(new_resource.aliases)
+#  end
 
   apache_site new_resource.site do
     cookbook "mediawiki"
diff --git a/cookbooks/mediawiki/templates/default/apache.erb b/cookbooks/mediawiki/templates/default/apache.erb
index d5485a62..36715815 100644
--- a/cookbooks/mediawiki/templates/default/apache.erb
+++ b/cookbooks/mediawiki/templates/default/apache.erb
@@ -23,9 +23,9 @@
 
   ServerAdmin webmaster@openstreetmap.org
 
-  SSLEngine on
-  SSLCertificateFile /etc/ssl/certs/<%= @name %>.pem
-  SSLCertificateKeyFile /etc/ssl/private/<%= @name %>.key
+#  SSLEngine on
+#  SSLCertificateFile /etc/ssl/certs/<%= @name %>.pem
+#  SSLCertificateKeyFile /etc/ssl/private/<%= @name %>.key
 
   CustomLog /var/log/apache2/<%= @name %>-secure-access.log combined
   ErrorLog /var/log/apache2/<%= @name %>-secure-error.log
```

migurski:
Great @mmd-osm! I expect @tomhughes to have an opinion on this. If I can get this cookbook running under Ubuntu without incident, I’d be willing to set up and fund a three-month experiment with EC2 to see if working tests support better contributions here.

mmd-osm:
Let's reflect for a moment about the ultimate goals of this thing:

>We have people who want to change the configuration of wiki.openstreetmap.org, and it would be easier for them if they could converge the recipes locally, and easier for admins if it was easy to check PRs locally before deploying them

I guess the EC2 instance wouldn't help with the first point (local testing). Regarding the second point, would an admin be able to log on to the EC2 instance and do a live test of those changes? I mean right now all we get back from test kitchen is some status that the build was successful. It won't tell anything about functional correctness:

Like in my example I removed SSL and an important Wiki extension to fight spam, still the overall test kitchen result was successful. Deploying such a change for sure would cause some major issues.

Which pieces are really missing here to turn this into something useful?

tomhughes:
What's missing is for you to all be quiet for a few days until I can investigate properly and find out what the real problems are and how to fix them!

I'm not really sure what EC2 has to do with anything, or how it would be used.

.kitchen.yml
```diff
@@ -10,7 +10,9 @@ provisioner:
   data_bags_path: test/data_bags
 
 platforms:
-  - name: ubuntu-16.04
+  # Ubuntu 18.04 is what’s used on the live OSM servers,
+  # according to https://hardware.openstreetmap.org/
+  - name: ubuntu-18.04
```
tomhughes: I updated this on master last week, so it shouldn't be needed here if you rebase.

cookbooks/accounts/attributes/default.rb
```diff
@@ -4,3 +4,6 @@
 
 default[:accounts][:groups] = {}
 default[:accounts][:users] = {}
+
+# Needed in cookbooks/accounts/recipes/default.rb to assign group membership
+default[:etc] = {"passwd": {}}
```
tomhughes: This is not really correct - the real problem is that you were using chef 15 which has the `Passwd` ohai plugin disabled by default. If we are ever able to switch to chef 15 then we would have to change that but for now the correct solution, which I have implemented on master, is to use chef 14 for the tests to match what we are running in production.

cookbooks/apt/recipes/default.rb
tomhughes: As already discussed this is already changed on master if you rebase.

migurski:
I’ll pick up the EC2 continuous testing conversation at some other point, to give you a little peace & quiet here Tom. :)

cookbooks/apt/recipes/default.rb
migurski: I merged in your upstream changes, thanks.

.kitchen.yml
migurski: Merged in upstream, thank you!

cookbooks/accounts/attributes/default.rb
migurski: 👍 I see your change in 21187d302463ac14ce12ea0b42d27d277d2f175a and I’m running it locally right now to verify.

migurski: Removed this in 5d218808c392f900121179e0d0c48c6ea5ce41b8 after testing.

migurski:
Cool, we’re back to the original error with less cruft in the PR: `1048 Column 'cuc_user' cannot be null`.

tomhughes:
So the `cuc_user` error is the only remaining issue as far as I can tell and as @mmd-osm says it comes from the `CheckUser` extension and specifically from https://github.com/wikimedia/mediawiki-extensions-CheckUser/blob/REL1_33/maintenance/populateCheckUserTable.php#L80 which runs as part of the installation to populate the initial data.

It is adding rows to `cu_changes` for each row in `recentchanges` which in this case just has a single row for the initial version of the main page as this is a new wiki - that has a user of zero which somehow turns into NULL for the insert, because because there is no user with an id of zero.

I assume it's just a bug in the extension or something, and would likely happen with a real deployment.

migurski:
I bet the extension is typically installed on already-populated wikis with existing user accounts, but the zero-vs.-null thing smells like MySQL lameness. Is this an upstream fix to CheckUser we need to suggest, or a potential fix by way of re-ordering the Mediawiki extensions?

mmd-osm:
Can you inject some new Mediawiki users via kitchen data bags so we don't have to deal with an empty database and a non-working default user?

I think it makes sense to also report this issue upstream on some wikimedia phabricator site, rather than trying to work out some feasible sequence to run those extensions, thereby only hiding the bug.

tomhughes:
Well injecting a user won't help unless whatever created that page record in the database uses it - there are already two users in the mediawiki database that the installer created, but they are users 1 and 2 and the page version was created against user 0.

So yes this probably needs reporting somwhere.

mmd-osm:
That's probably here, then: https://phabricator.wikimedia.org/maniphest/task/edit/form/1/?projects=CheckUser

(project page for the extension: https://www.mediawiki.org/wiki/Extension:CheckUser)

migurski:
Thought I’d re-ask my question from a few weeks back: I think I’ve made the requested changes, and this PR incrementally improves the repo even if the wiki cookbook does not yet fully complete. Can we merge this work, for closure and turn the CheckUser work into a new issue?

mmd-osm:
@migurski : could you possibly do a follow up with CheckUser folks to get this resolved while we keep this issue open?

I don't quite see the benefit in rushing this PR. It would only increase the frustration of people eager to try this out, only to find out that they're hitting roadblocks they cannot solve on their own. Non-working stuff also increases support burden.

migurski:
I don’t know if I agree that closing after 2 1/2 weeks is rushing? From my experience, dangling open PRs are a nuisance for potential contributors who encounter a repo and can’t reason about its present status. Closure here would do a better job of communicating what’s done (a handful of data bags plus the actual fixes that @tomhughes separately cherry-picked into master) and what isn’t (an upstream Mediawiki dependency that also wasn’t working previously).

I do not currently have the bandwidth to follow up on CheckUser. There might be someone else who could do it without needing to also understand the long comment history here. I’d like to see this PR closed.

mmd-osm:
> are a nuisance for potential contributors who encounter a repo and can’t reason about its present status. 

If you're concerned about communicating the status, I'd suggest Github tagging and flag the issue as "waitfor-upstream". iD editor uses this, and I think it does a good job.

migurski:
Thanks @mmd-osm, that’s a new convention to me and I wasn’t previously familiar with it. Is it something used here, or just in iD?

mmd-osm:
iD has a fair amount of labels in place to categorize issues. All "wait-*" labels mark issues/pull request where an external reaction/decision is pending: https://github.com/openstreetmap/iD/labels?utf8=%E2%9C%93&q=waitfor  You can filter issues by tags, add/remove them as the issue progresses.

It's probably not in used around here (yet).

tomhughes:
I've raised https://phabricator.wikimedia.org/T244283.

migurski:
Wonderful! In case it’s of use, here’s a small patch to CheckUser I tested that appears to circumvent our problem: https://github.com/migurski/extension-checkuser/commit/34c693295b9092b5954562eb68ead92d995ab7d5 

I don’t know enough about MW or CU to guess whether this is correct, but it does suppress the NULL insert.

migurski:
With the changes from #266 on master, I’ve rebased this PR to expose the wiki to testing on Travis. I’ve worked around the CheckUser problem here by referencing a change to a fork of the plugin (https://github.com/migurski/extension-checkuser/commit/96c0441900b83e179306994b0fac8dc866e831c7).

Firefishy:
:fireworks: IT PASSED! https://travis-ci.org/openstreetmap/chef/jobs/650967907

migurski:
It is a beast of a test! 

The funky CheckUser fork might be a cause for concern, though I see there are other instances of MW plugins that we host ourselves. Curious to see if MW accepts our changes.

tomhughes:
@Firefishy I'm waiting to see what happens on the mediawiki bug - I'd prefer not to switch to a fork and I definitely don't want to deploy this without some indication from upstream that the change is not going to break things.

migurski:
I’m continuing to rebase as-needed while waiting for follow up on https://phabricator.wikimedia.org/T244283

.travis.yml
```diff
@@ -9,10 +9,22 @@ jobs:
...
+    - name: "Test Kitchen (web::backend)"
+      script:
+        - bundle exec kitchen test web-backend-ubuntu-1804
+    - name: "Test Kitchen (web::frontend)"
+      script:
+        - bundle exec kitchen test web-frontend-ubuntu-1804
+    - name: "Test Kitchen (web::rails)"
+      script:
+        - bundle exec kitchen test web-rails-ubuntu-1804
     - name: "Test Kitchen (wiki)"
```
tomhughes: Please don't mix unrelated changes in one pull request - stick to making the wiki changes in this PR and if you want to propose reordering the tests then do that separately.

migurski:
MW closed https://phabricator.wikimedia.org/T244283 earlier this morning so I’ve removed the fork of CheckUser and rebased this PR to re-run tests. The wiki test runs clean now!

I looked at each of the other test failures, and all of them except one fail due to the same regrettable flaky `git.openstreetmap.org` checkout: 

    Failed to connect to git.openstreetmap.org port 443: Connection timed out

The [civiccrm test failure](https://github.com/openstreetmap/chef/pull/259/checks?check_run_id=515555557) looks like a failure to check out https://github.com/veda-consulting/uk.co.vedaconsulting.mailchimp/commit/25798ffe209650f12162a0cdc3c7ee9203d950f2 though the ref named in the attributes does still exist: https://github.com/openstreetmap/chef/blob/8fe15ddbca7a39536f23303076bd8f4591128a78/cookbooks/civicrm/attributes/default.rb#L23-L25


Might be worth bringing that Mailchimp plugin up to a newer commit or master. [Here’s what’s changed since `25798ffe`](https://github.com/veda-consulting/uk.co.vedaconsulting.mailchimp/compare/25798ffe2096..master).

mmd-osm:
Regarding the timeout issues we're seeing with Github Actions, I would actually raise an issue with Github Actions team for further analysis. Maybe we're hitting some kind of limit?

This doesn't only affect `git.openstreetmap.org` but also also  fetching keys from Ubuntu keyservers, and some other external sites that are being queried in some of the cookbooks. All of them bail out with a timeout once in a while.

Moving git.openstreetmap.org to another DC won't be sufficient to address this issue in that case.

tomhughes:
I believe @grischard is already working on making contact.

mmd-osm:
Ok, sounds good. I've read this bit here on https://wiki.osmfoundation.org/wiki/Operations_Working_Group/Minutes/2020-04-10 and thought that there might be a bit of root cause analysis needed.

> Fixing the failing CI tests
> 
> Issue: Unreliable https://github.com/openstreetmap/chef/actions
> Should we make Github Primary? Should we move git.openstreetmap.org to another data centre?

PR #260, opened 2020-01-15
Updated Mediawiki apt repo signing key to current value

migurski:
https://www.mediawiki.org/wiki/Parsoid/Setup#Ubuntu_/_Debian indicates that "AF380A3036A03444" is the current key. "90E9F83F22250DD7" did not work in testing.

Following suggestion in https://github.com/openstreetmap/chef/pull/259#discussion_r367044857


migurski:
Nm, I see this has already been changed.

PR #262, opened 2020-01-27
planet mirror: also add redirect for planet.pbf

mtmail:
fixes https://github.com/openstreetmap/operations/issues/355

Regexp is only checked for match, individual parts are not used thus adding another bracket was fine.


mmd-osm:
Full history files were also included in the manual redirect: e.g. https://planet.openstreetmap.org/pbf/full-history/history-200120.osm.pbf

mtmail:
xml and pdf paths are different so my change wouldn't work.

PR #266, opened 2020-02-09
Configured Travis CI to run cookbooks via Test Kitchen

migurski:
In https://github.com/openstreetmap/chef/pull/259, we noted that observing Test Kitchen outputs required manual sharing of logs and some conversation about working Test Kitchen installations. This PR introduces the possibility of testing Chef cookbooks right in Travis CI, thanks to the use of [Docker under Kitchen-Dokken](https://github.com/someara/kitchen-dokken) which will successfully complete in a virtual environment such as Travis.

@jalessio and I have started with the small Apt cookbook suite as a proof of concept. Cookstyle and Kitchen tests now run in parallel with logs that can be shared and observed between editors in PRs to this repo. The [end-user display in Travis is pretty nice](https://travis-ci.org/openstreetmap/chef/builds/648120451):

Screen Shot 2020-02-09 at 12 08 11 PM

If accepted, future work would include adding each cookbook suite (Wiki, Website, etc.) in turn to the Travis config and ensuring that it passes tests. @gravitystorm has recommended that we break this work up over time and add remaining suites in followup PRs.

Notes:
- We’ve defined the Kitchen-Dokken platform somewhat redundantly to leave the current Vagrant default untouched. A future version of `.kitchen.yml` could be simplified a lot by abandoning Vagrant entirely, but for now this change should be completely backwards-compatible with existing Vagrant setups.
- Current Travis time limitation is 50 minutes per matrix build job. Optimistically this should cover all of our needs, but as we add Cookbook tests we may find that some of them run very slowly. We can try to increase our runtime ceiling with a paid Travis plan or use an alternate Test Kitchen driver such as EC2 if we encounter this problem in the future.
- Any Kitchen command with the current Vagrant driver present requires that `vagrant --version` returns a valid version string. Since we [can’t actually run Vagrant/Virtualbox on Travis](https://github.com/travis-ci/travis-ci/issues/6060#issuecomment-221395944), we cheat a little here by generating a one-line `/usr/bin/vagrant` to echo a fake version string. We could also `apt-get install vagrant` to be more correct, with some startup time cost.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -22,6 +22,7 @@ platforms:
     driver:
       name: dokken
       image: dokken/ubuntu-18.04
+      chef_version: 14
       provision: true
       # Docker and systemd need full capabilities granted in privileged mode
       privileged: true
@@ -33,10 +34,7 @@ platforms:
     provisioner:
       name: dokken
       product_name: chef
-      product_version: 14
       data_bags_path: test/data_bags
-      # Automatically accept Chef license when running headlessly
-      chef_license: accept
     transport:
       name: dokken
 
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -12,20 +12,30 @@ provisioner:
   data_bags_path: test/data_bags
 
 platforms:
+  # Ubuntu 18.04 is what’s used on the live OSM servers,
+  # according to https://hardware.openstreetmap.org/
   - name: ubuntu-18.04
   - name: ubuntu-18.04-ci
+    # Driver, transport, and provisioner using Docker and Chef Infra Client,
+    # necessary when used on virtual CI boxes that refuse to run Virtualbox or Vagrant.
+    # https://github.com/someara/kitchen-dokken
     driver:
       name: dokken
       image: dokken/ubuntu-18.04
       provision: true
+      # Docker and systemd need full capabilities granted in privileged mode
+      privileged: true
+      # Run systemd init to allow services like MySQL, etc. to function
       pid_one_command: /bin/systemd
       intermediate_instructions:
+        # Help APT find its required sources on first run
         - RUN /usr/bin/apt-get update -y
     provisioner:
       name: dokken
       product_name: chef
       product_version: 14
       data_bags_path: test/data_bags
+      # Automatically accept Chef license when running headlessly
       chef_license: accept
     transport:
       name: dokken
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -13,28 +13,49 @@ provisioner:
 
 platforms:
   - name: ubuntu-18.04
+  - name: ubuntu-18.04-ci
+    driver:
+      name: dokken
+      image: dokken/ubuntu-18.04
+      provision: true
+      pid_one_command: /bin/systemd
+      intermediate_instructions:
+        - RUN /usr/bin/apt-get update -y
+    provisioner:
+      name: dokken
+      product_name: chef
+      product_version: 14
+      data_bags_path: test/data_bags
+      chef_license: accept
+    transport:
+      name: dokken
 
 suites:
   - name: accounts
     run_list:
       - recipe[accounts::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: apache
     run_list:
       - recipe[apache::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: apt
     run_list:
       - recipe[apt::default]
   - name: bind
     run_list:
       - recipe[bind::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: blogs
     run_list:
       - recipe[accounts::default]
       - recipe[blogs::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: forum
     run_list:
       - recipe[accounts::default]
       - role[forum]
+    excludes: ["ubuntu-18.04-ci"]
   - name: letsencrypt
     run_list:
       - recipe[accounts::default]
@@ -44,26 +65,34 @@ suites:
       apt:
         sources:
           - openstreetmap
+    excludes: ["ubuntu-18.04-ci"]
   - name: munin
     run_list:
       - recipe[munin::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: munin-server
     run_list:
       - recipe[munin::server]
+    excludes: ["ubuntu-18.04-ci"]
   - name: mysql
     run_list:
       - recipe[mysql::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: networking
     run_list:
       - recipe[networking::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: otrs
     run_list:
       - recipe[accounts::default]
       - recipe[chef::default]
       - role[otrs]
+    excludes: ["ubuntu-18.04-ci"]
   - name: python
     run_list:
       - recipe[python::default]
+    excludes: ["ubuntu-18.04-ci"]
   - name: tools
     run_list:
       - recipe[tools::default]
+    excludes: ["ubuntu-18.04-ci"]
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,20 @@
 sudo: false
 language: ruby
 cache: bundler
-script:
-  - bundle exec cookstyle -f fuubar
+services:
+  - docker
+before_script:
+  - gem install kitchen kitchen-dokken kitchen-vagrant
+  # Put just enough Vagrant in the PATH to satisfy Test-Kitchen. No actual
+  # Vagrant or Virtualbox is run on Travis; this fake shim that echoes a
+  # recent Vagrant version is only here to allow tests to proceed below.
+  - sudo bash -c 'echo "echo Vagrant 2.0.0-fake" > /usr/bin/vagrant'
+  - sudo chmod +x /usr/bin/vagrant
+jobs:
+  include:
+    - name: "Cookstyle"
+      script:
+      - bundle exec cookstyle -f fuubar
+    - name: "Kitchen Test Apt Cookbook"
+      script:
+      - kitchen test apt-ubuntu-1804-ci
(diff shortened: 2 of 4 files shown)


tomhughes:
This rather overlaps with the work I've been going to try and run the tests under docker but as expected my experience is that not all of them can be run under docker so unless you have a magic solution for that I'm not sure this is a feasible approach?

tomhughes:
Incidentally I've never heard of the "dokken" driver - what made you chose that over the standard docker driver?

migurski:
Andy recommended the Dokken driver: it’s designed to emulate complete systems, has a functioning init, and is built specifically for Test Kitchen and Chef. @jalessio and I went some distance down the path of the plain Docker driver and found that we were duplicating existing work from Dokken.

jalessio:
> my experience is that not all of them can be run under docker so unless you have a magic solution for that I'm not sure this is a feasible approach?

@tomhughes I can't definitively say _yet_ that all of these will run under Docker, but one of the main issues I've encountered while testing Chef cookbooks via Docker is that you often need an init system running in order to test daemons such as MySQL. We got over this hurdle by utilizing the line `pid_one_command: /bin/systemd` _(which also requires the addition of `privileged: true`)_. With this in place you can start multiple background processes (and make assertions against them). This unlocks a lot of testing possibilities with Docker + Chef + Test Kitchen.

I have a branch which uses this code to test the mysql cookbook from this repo, which demonstrates that the MySQL daemon starts successfully as part of the CI suite using Docker on Travis. It's part of this build (https://travis-ci.org/jalessio/openstreetmap-chef/builds/648127647) and sample MySQL logs are here (https://travis-ci.org/jalessio/openstreetmap-chef/jobs/648127652).

My [kitchen-add-passing-cookbooks-to-ci](https://github.com/jalessio/openstreetmap-chef/tree/kitchen-add-passing-cookbooks-to-ci) branch currently has builds working on Travis for 6 of the 14 cookbooks in this repo. On that branch, I **only** added cookbooks which passed without any modification _(as an easy starting point)_.
Screen Shot 2020-02-09 at 1 07 47 PM
_(screenshot from https://travis-ci.org/jalessio/openstreetmap-chef/builds/648127647)_

I haven't dug into each of the failing cookbooks yet, but I don't think the failures are necessarily related to using Docker since they are failing for reasons like missing test Data Bags _(meaning they probably fail under Vagrant too)_.

As for utilizing Dokken, it does take care of some things we'd have to build anyway to make this work and my confidence was boosted when I saw it being [used extensively in the check-cookbooks repo](https://github.com/chef-cookbooks/docker/blob/master/kitchen.yml).

tomhughes:
Yes the init system is the main issue but I did actually have that running and still had some issues. Plus I was actually using `podman` and I wasn't sure real docker was up to running systemd as pid one.

I'll try and have a look at this at some point this week.

tomhughes:
Is there a reason why you have added `kitchen-dokken` as a separate platform rather than just switching to it?

I'd rather not have the complexity of supporting two different platforms for the tests if we can avoid it, and it we can make the tests run in a container environment then that would seem to be the better option from a performance point of view.

Personally I would have preferred it if this PR was split in two so that we could address the question of which driver to use separately from the question of running the tests in travis - aside from anything else I'm not sure it's actually practical to run the tests in travis once we have more testing just because of the length of time it would take.

I've had a brief look at `kitchen-dokken` and I can see how it speeds things up more - sadly it looks like it works via the docker API so requires real docker with a daemon which is less than ideal for me personally but that's not a deal breaker.

migurski:
Re: separate platform, Jamie and I had set up Docker as a complete Vagrant replacement originally and then looked for a way to make this change less invasive. This PR introduces Docker specifically for use in CI without upending any existing Vagrant setup you or anyone else from the OWG or sysadmins teams might have. A future version of `.kitchen.yml` could be simplified a lot by abandoning Vagrant entirely but it seemed very drastic for this proof of concept.

Re: split PR, I’m unsure it’s possible to split these two ideas. Only a Docker-based driver will work on hosted CI like our Travis integration. We previously tried to make this change work without a new driver but I think these two modifications have to go hand-in-hand. It’s possible to address the length of time by [increasing our Travis parallelism with a paid plan](https://travis-ci.com/plans): most of these test suites are quite speedy.

Glad to hear Docker is not a deal-breaker for you!

tomhughes:
The thing is that I'm reluctant to commit to a specific technology until we know it will work across the board, which is why I'd rather look at switching technology to something that will work in travis first.

Once I've reinstalled docker here I'll have a go with this and see how it goes across the whole range of test suites to get some idea of what the status is. I have my own notes from trying kitchen-docker as a baseline to compare to.

tomhughes:
Unfortunately it turns out that Fedora 31 can't run docker anymore, at least not without rebooting to go back to groupsv1, so I'm not going to be able to test this right now.

.kitchen.yml
```diff
@@ -12,29 +12,60 @@ provisioner:
...
+      intermediate_instructions:
+        # Help APT find its required sources on first run
+        - RUN /usr/bin/apt-get update -y
+    provisioner:
+      name: dokken
+      product_name: chef
+      product_version: 14
+      data_bags_path: test/data_bags
+      # Automatically accept Chef license when running headlessly
+      chef_license: accept
```
tas50: if you're using chef 14 there's no need to accept the license

tomhughes: Yes I was going to point that out ;-) That said my first attempt at running it is trying to use 15 for some reason...

tas50:
Kitchen-dokken is our preferred method of testing chef cookbooks in CI systems. Kitchen-docker's PID 1 won't allow you to test most services. Kitchen-dokken will properly allow starting services. That coupled with the dokken images, which are "fat" containers allowing for full service testing, and you have fairly complete testing setup that can run in most CI environments.

.kitchen.yml
migurski: Interesting! We added this because we saw license-related failures until we added this line.

tas50: You're seeing that because this config isn't actually correct. product_version is for omnibus installs in kitchen-vagrant. Kitchen-dokken is fast in part because it just mounts the chef docker container onto the OS container image. Due to that you specify the chef version a bit differently. See the example here: https://github.com/someara/kitchen-dokken#usage

tas50: @tomhughes Feel free to fill out https://www.chef.io/nonprofit-entitlement-program/ if you want to use Chef 15.

Firefishy: Chef 14 fix for me: 
```
diff --git a/.kitchen.yml b/.kitchen.yml
index 449f56dc..fe3e165d 100644
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -22,6 +22,7 @@ platforms:
     driver:
       name: dokken
       image: dokken/ubuntu-18.04
+      chef_version: 14
       provision: true
       # Docker and systemd need full capabilities granted in privileged mode
       privileged: true
@@ -33,10 +34,7 @@ platforms:
     provisioner:
       name: dokken
       product_name: chef
-      product_version: 14
-      # Automatically accept Chef license when running headlessly
-      chef_license: accept
     transport:
       name: dokken
```

jalessio: @firefishy I’m curious - what problem occurs for you with these lines present? 

Certainly seems like removing them is the right thing to do, but it’s confusing to me since neither me or @migurski ran into any issues with these in place.

tomhughes: @jalessio The problem is that you are actually running with chef 15 not chef 14 and although the test you chose works under chef 15 not all of them do, and chef 14 is what we use in production right not and what you should testing.

tomhughes: https://github.com/tomhughes/openstreetmap-chef/commit/2744cdd0988f14d5bf2cf9c9563b3fb96d60bf61 is the configuration I am currently testing with - most of the tests are working and I am just fixing up some long standing issues that affect even our current environment.

.kitchen.yml
```diff
@@ -12,29 +12,60 @@ provisioner:
...
+  - name: ubuntu-18.04-ci
+    # Driver, transport, and provisioner using Docker and Chef Infra Client,
+    # necessary when used on virtual CI boxes that refuse to run Virtualbox or Vagrant.
+    # https://github.com/someara/kitchen-dokken
+    driver:
+      name: dokken
+      image: dokken/ubuntu-18.04
+      provision: true
+      # Docker and systemd need full capabilities granted in privileged mode
+      privileged: true
```
mmd-osm: I tried dokken/ubuntu-18.04 with privileged: true and it always seemed to interfere with my local system, i.e. logging off users, interfering with X server, etc. Curious if you experience something similar when running this locally?

tomhughes: I'm running in an ssh session on errol and haven't noticed anything odd, but there's no X or anything there.

migurski: I haven’t experienced anything like this when testing on Mac OS X 10.14.

.kitchen.yml
migurski: These are great suggestions, thanks for the detail and hints @tas50.

@tomhughes it looks like you’re constructing your own version of our branch in parallel, would you prefer to take over this PR?

tomhughes:
I believe I've now fixed all the outstanding issues with the tests, and have a configuration at https://github.com/tomhughes/openstreetmap-chef/tree/dokken that can run them all in docker and in travis, as seen here:

https://travis-ci.org/tomhughes/openstreetmap-chef/builds/649655674

migurski:
Exciting! Do you plan to accept this compatible PR which includes comments in the config files explaining how things work, or would you like for us to close #266 out so you can use the version you’ve written?

tomhughes:
I took the commits from this PR and rearranged them slightly then added my work on top so that preserve as best as possible attribution for the work - that is all now pushed on master.

The only thing I haven't taken is the comments - many of them were either no longer relevant or were just documenting stuff that was boilerplate from the dokken instructions so I left them out.

Hopefully that all works for you, and thanks for your work on this, and for finding dokken!

migurski:
This works, and I’m excited to see new expanded test coverage on master!

PR #269, opened 2020-02-15
blogs: use bundle install deployment over global

Firefishy:
Switch to using safe `bundle install --deployment` instead of
global install, which may conflict with other dependencies.

### DIFF
--- a/cookbooks/blogs/recipes/default.rb
+++ b/cookbooks/blogs/recipes/default.rb
@@ -43,6 +43,7 @@ end
 git "/srv/blogs.openstreetmap.org" do
   action :sync
   repository "git://github.com/gravitystorm/blogs.osm.org.git"
+  depth 5
   user "blogs"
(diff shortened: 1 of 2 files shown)


cookbooks/blogs/recipes/default.rb
```diff
@@ -43,14 +43,15 @@
...
+  depth 5
   user "blogs"
   group "blogs"
   notifies :run, "execute[/srv/blogs.openstreetmap.org/Gemfile]", :immediately
 end
 
 execute "/srv/blogs.openstreetmap.org/Gemfile" do
   action :nothing
-  command "bundle install"
+  command "bundle install --deployment"
```
tomhughes: If you're doing this then it can run as user `blobs` instead of root.

Firefishy: Done.

PR #270, opened 2020-02-18
Re-ordered tests to minimize Travis wall time

migurski:
Lengthy web-related tests go up-front, remaining short tests come later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,19 @@ jobs:
     - name: "Cookstyle"
       script:
         - bundle exec cookstyle -f fuubar
+
+    # Selected long-running tests are at the top to minimize Travis wall time
+    - name: "Test Kitchen (web::backend)"
+      script:
+        - bundle exec kitchen test web-backend-ubuntu-1804
+    - name: "Test Kitchen (web::frontend)"
+      script:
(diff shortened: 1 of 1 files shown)


tomhughes:
I understand the logic but I'm reluctant to break the readability of having the tests in order for this.

On top of that it's not clear that we're going to be staying with travis-ci as our CI solution given the limitations that you are trying to workaround here so this may not be necessary depending on what decisions we make.

migurski:
Cool. I’ll close this out while you decide on a CI strategy for the future.

tas50:
@tomhughes It would be pretty easy to switch this all over to Github Actions which would greatly speed up executing the tests.

tomhughes:
Like https://github.com/tomhughes/openstreetmap-chef/commit/c0371f73093f8d07306d7ebffb77775e2447e0c2 you mean ;-)

Example run at https://github.com/tomhughes/openstreetmap-chef/actions/runs/41230861.

tas50:
@tomhughes Totally like that. You might want to check out some of the canned actions in https://github.com/actionshub/. We're writing those in Sous Chefs for testing all those cookbooks and they've been pretty handy.

PR #278, opened 2020-02-23
Add new tile layers to MultiMaps extension

Tigerfell:
Then replacing Slippy Map with MultiMaps extension, I found that the first supports displaying Thunderforest's Cycle layer, while MultiMaps was not configured to do that. 
Additionally, I saw that a lot of French pages included links to the French style tile server, that is why I suggest to add this layer, too. This was previously discussed in cquest/osmfr-cartocss/issues/61 and osm-fr/infrastructure/issues/153.

### DIFF
--- a/cookbooks/wiki/templates/default/mw-ext-MultiMaps.inc.php.erb
+++ b/cookbooks/wiki/templates/default/mw-ext-MultiMaps.inc.php.erb
@@ -14,4 +14,14 @@ $egMultiMaps_MapServices = [
     'attribution' => '&copy; <a href="https://osm.org/copyright">OpenStreetMap contributors</a>. Tiles courtesy of <a href="https://www.thunderforest.com/" target="_blank">Andy Allan</a>',
     'source' => 'https://tile.thunderforest.com/transport/{z}/{x}/{y}.png?apikey=<%= @thunderforest_key %>',
   ],
+  'cycle' => [
+    'service' => 'Leaflet',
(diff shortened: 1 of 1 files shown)


tomhughes:
I see no problem with the cycle layer but I'm not sure policy wise whether we should limit it to layers that appear on the main site or whether we should allow additional layers like the french one that aren't on the main site?

I would also just point out that it was not at all clear from either the title or the description what this PR was about or I might have looked at it before...

Tigerfell:
I guess you refer to [Guidelines for new tile layers](https://wiki.openstreetmap.org/wiki/Featured_tile_layers/Guidelines_for_new_tile_layers) wiki article which is not listed at https://operations.osmfoundation.org/policies/ (meaning that anyone could have modified the policy without agreement). 
French style meets all required criteria and at least soft criteria "interesting" (due to experimental features), "non-comercial", and "free and open". I guess it would be supportive for OSM-FR community to see that their layer is available in the wiki.

tomhughes:
Right, but meeting the criteria is the first step - then somebody has to propose it be added and the OWG votes on it.

That is for adding a layer to the main site however - my point here was whether we should keep the layers in the wiki in sync with the main site or allow additional layers.

Tigerfell:
Humanitarian layer was never available in the wiki and there is no necessity for it, because HOT wiki pages usually do not feature map displays. 
The only case where I could imagine some synchronization is when there is a request for a new tile layer for osm.org and you would ask the requester if they want to have it in the wiki, too. I do not see a problem with either site having additional layers.

Tigerfell:
Superseded by #331

PR #287, opened 2020-03-21
Add Nginx totals to tile.openstreetmap.org munin

pnorman:
We have nginx per-cache, but no total graph.

I'm not sure how I can test this locally, is there any sensible way?


pnorman:
Yay, got it right :)

PR #288, opened 2020-03-24
Add fake MS referer to block list

pnorman:
These are all coming with a referer of `Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.2; .NET CLR 1.0.3705;)` so we could block that instead

### DIFF
--- a/cookbooks/tilecache/templates/default/nginx_tile.conf.erb
+++ b/cookbooks/tilecache/templates/default/nginx_tile.conf.erb
@@ -93,6 +93,7 @@ map $http_referer $denied_referer {
   'https://www.google.com'         1; # Faked
   'https://google.com/'            1; # Faked
   'https://www.google.com/'        1; # Faked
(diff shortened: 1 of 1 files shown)


pnorman:
Most of the requests with this fake referer are also coming from two IPs, so we could block them that way too

tomhughes:
Are we sure that's bogus? It's coming back as IE6 on Server 2003 when I analyse it... or do we just think there is an implausible volume for such an old platform?

pnorman:
> Are we sure that's bogus? It's coming back as IE6 on Server 2003 when I analyse it... or do we just think there is an implausible volume for such an old platform?

A quick grepping of logs showed all the requests with that UA had the fake referer. I didn't check for 100%, but this is blocking the referer which definitely is fake.

tomhughes:
Oh sorry I missed that it was the referer. That is of course insane...

PR #289, opened 2020-03-25
imagery: reprojected lu hillshade

DavidMoraisFerreira:
Hi,
The hillshade tif was re-projected server-side, the offset visible [here](https://ana-dtm-2017.openstreetmap.lu/) should be solved.
@Firefishy, @grischard : gdaladdo with --config PHOTOMETRIC_OVERVIEW YCBCR failed with 

> ERROR 6: PHOTOMETRIC_OVERVIEW=YCBCR requires a source raster with only 3 bands (RGB)

, so I used the default value.

Edit: the commands I used are stored in the file `commands_executed.sh` on the server.


grischard:
Merged in dbde4095

PR #291, opened 2020-03-28
Add email for MWG osmf-talk owner queue

pnorman:
The OTRS side is set up

### DIFF
--- a/roles/otrs.rb
+++ b/roles/otrs.rb
@@ -32,6 +32,15 @@ default_attributes(
         :group => "www-data",
         :home_directory => "/opt/otrs"
       },
+      :otrs_membership_osmf_talk_owner => {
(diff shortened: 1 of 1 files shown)


tomhughes:
Sorry what's this for exactly?

pnorman:
It's to let OTRS pick up from the email osmf-talk-owner@otrs.openstreetmap.org, which will go into its own dedicated queue. We'll then switch over the email in mailman, so instead of mailing list stuff ending in the general MWG queue, it'll have its own queue that it automatically goes into.

tomhughes:
Blimey, how much email does osmt-talk-owner get that's it worth all this trouble?

pnorman:
For form submissions we can sort into queues based on the email it comes from, but the only way I know of to get going into the right queue working is with its own email

PR #293, opened 2020-04-06
Create gackelchen, gandi and lu roles

grischard:


roles/gackelchen.rb
```diff
@@ -0,0 +1,46 @@
...
+    :cache_mem => "20480 MB",
+    :cache_dir => [
+      "rock /store/squid/rock-4096 20000 swap-timeout=200 slot-size=4096 max-size=3996",
+      "rock /store/squid/rock-8192 25000 swap-timeout=200 slot-size=8192 min-size=3997 max-size=8092",
+      "rock /store/squid/rock-16384 35000 swap-timeout=200 slot-size=16384 min-size=8093 max-size=16284",
+      "rock /store/squid/rock-32768 45000 swap-timeout=200 slot-size=32768 min-size=16285 max-size=262144"
+    ]
+  },
+  :tilecache => {
+    :tile_parent => "netherlands.render.openstreetmap.org"
```
tomhughes, line 39: That needs to be changed to match the name of the tile CDN node - normally the name of the city but sometimes the country if it's a multi-noce cache. In this case `bissen.render.openstreetmap.org` I think.

roles/gandi.rb
```diff
@@ -0,0 +1,24 @@
+name "gandi"
+description "Role applied to all servers at Gandi"
+
+default_attributes(
+  :hosted_by => "Gandi",
+  :location => "Bissen, Luxembourg",
+  :networking => {
+    :roles => {
+      :external => {
+        :zone => "gandi"
```
tomhughes, line 10: Just use `osm` as the zone - some of the older nodes do have a provider specific zone but I'm no longer doing that other than for the big sites with lots of machines and if you do add a new zone like that then the networking cookbook needs changing to define it.

roles/gandi.rb
```diff
@@ -0,0 +1,24 @@
+name "gandi"
+description "Role applied to all servers at Gandi"
+
+default_attributes(
+  :hosted_by => "Gandi",
+  :location => "Bissen, Luxembourg",
+  :networking => {
+    :roles => {
```
tomhughes, line 8: There should be something here setting the nameservers to use.

grischard: Gandi's server management stuff actually sets this itself, and the servers come pre-configured with systemd-resolvd.

```
root@gackelchen:~# systemd-resolve --status
Global
         DNS Servers: 217.70.186.194
                      217.70.186.193
                      2001:4b98:dc2:49::193
          DNSSEC NTA: 10.in-addr.arpa
                      16.172.in-addr.arpa
                      168.192.in-addr.arpa
                      17.172.in-addr.arpa
                      18.172.in-addr.arpa
                      19.172.in-addr.arpa
                      20.172.in-addr.arpa
                      21.172.in-addr.arpa
                      22.172.in-addr.arpa
                      23.172.in-addr.arpa
                      24.172.in-addr.arpa
                      25.172.in-addr.arpa
                      26.172.in-addr.arpa
                      27.172.in-addr.arpa
                      28.172.in-addr.arpa
                      29.172.in-addr.arpa
                      30.172.in-addr.arpa
                      31.172.in-addr.arpa
                      corp
                      d.f.ip6.arpa
                      home
                      internal
                      intranet
                      lan
                      local
                      private
                      test

Link 2 (eth0)
      Current Scopes: none
       LLMNR setting: yes
MulticastDNS setting: no
      DNSSEC setting: no
    DNSSEC supported: no
```

It also comes preconfigured with systemd-timesyncd for ntp:

```
root@gackelchen:~# systemctl status systemd-timesyncd
● systemd-timesyncd.service - Network Time Synchronization
   Loaded: loaded (/lib/systemd/system/systemd-timesyncd.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2020-04-06 15:59:53 CEST; 2h 32min ago
     Docs: man:systemd-timesyncd.service(8)
 Main PID: 934 (systemd-timesyn)
   Status: "Synchronized to time server 91.189.89.198:123 (ntp.ubuntu.com)."
    Tasks: 2 (limit: 4915)
   CGroup: /system.slice/systemd-timesyncd.service
           └─934 /lib/systemd/systemd-timesyncd

Apr 06 15:59:53 gackelchen systemd[1]: Starting Network Time Synchronization...
Apr 06 15:59:53 gackelchen systemd[1]: Started Network Time Synchronization.
Apr 06 15:59:53 gackelchen systemd-timesyncd[934]: Synchronized to time server 91.189.89.198:123 (ntp.ubuntu.com).
```

tomhughes: That won't survive a chef run and reboot though - either that is statically configured and will be overwritten by chef or it came from DHCP and will therefore no longer apply once chef has switched the machine to be statically configured.

tomhughes: Oh and chef will turn off systemd-timesyncd in favour of chrony.

PR #299, opened 2020-04-29
RFC-formatted complainant address for easy copy/paste

polarbearing:
Add fully formatted complainant address to dmca mail body for easy copy/pasting in responses
as discussed in the dwg. 
( $reply_address can potentially be used for a Reply-To header )


polarbearing:
What should I do about these fails? Thing like ```Error executing action `create` on resource 'rails_port[www.openstreetmap.org]'``` appear to be unrelated to the 2 php lines I added?

tomhughes:
You let us worry about them.

PR #305, opened 2020-05-20
Add a bundler vendor cache

Firefishy:
DO NOT MERGE YET.

### DIFF
--- a/cookbooks/stateofthemap/recipes/default.rb
+++ b/cookbooks/stateofthemap/recipes/default.rb
@@ -312,6 +312,12 @@ gem_package "bundler" do
   version "2.1.4"
 end
 
+directory "/srv/stateofthemap.org-vendor-bundle" do
+  owner "root"
+  group "root"
+  mode 0o755
+end
+
 %w[2016 2017 2018 2019 2020].each do |year|
   git "/srv/#{year}.stateofthemap.org" do
     action :sync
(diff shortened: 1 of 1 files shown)


tomhughes:
No. Just no. Never.

Vendor if you want, but don't even think about committing the vendored modules.

tomhughes:
What on earth are you trying to do? The whole point of vendoring is to give each checkout it's own module set - if you want to share it then just use the global modules!

Firefishy:
> What on earth are you trying to do? The whole point of vendoring is to give each checkout it's own module set - if you want to share it then just use the global modules!

The `--deployment` does a few more things than purely vendoring. Capistrano symlinks `vendor/bundler` between releases to improve deployment speed, bundler is able to cope with the shared folder. I am experimenting to see if there is any real benefit.

HolgerJeromin:
> DO NOT MERGE YET.

We could use [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) in such a case.

mmd-osm:
Once thing I didn't understand with the stateofthemap jobs is that we don't use Jekyll that is provided by Ubuntu 18.04 out-of-the-box. Probably I'm missing something here.

A lot of time in this step is spent compiling some libsass library in C on a single CPU for each single year.

tomhughes:
Well because we want to make sure we use the version of jekyll specified in the bundle, otherwise we might not get the results the author of the site expected.

Firefishy:
Around 6 minutes is all we'd save. Likely better to find a way to parallelise the operations.

PR #306, opened 2020-05-29
Increase NIC rx/tx ring size

Firefishy:
Use udev to set the max supported rx/tx ring size for optimal NIC offload.


Firefishy:
I have not yet merged this because on 1 of the DL360 G6 machines I tested on the NIC locked and didn't cleanly come back after changing the ring size. Downing / Uping the NIC fixed the issue.

Firefishy:
Merged in 7c46dd134e9ebaa741a1a74e75b740be82ee0c49

PR #312, opened 2020-06-21
kitchen: Upgrade ca-certificates prior to run

Firefishy:
Upgrade `ca-certificates` to ensure current trusted root are installed.


tomhughes:
Nice try, but it's the bundle in the chef omnibus package that is the problem, along with the ancient version of openssl chef is using that stops it finding the alternate root that hasn't expired.

PR #325, opened 2020-07-22
Move OpenLayers to chef control + add custom OpenLayers 404.png

Firefishy:
This will need a PR to remove OpenLayers from openstreetmap-website repo.


tomhughes:
There's no actual requirement to remove it from the rails code given that we serve this in preference though obviously we can.

PR #326, opened 2020-07-22
Switch to un-blessed osmdata output

JesseWeinstein:
Due to a complicated and unfortunate mess, the blessed/marked good versions of the osmdata.openstreetmap.de output files have not been updated since Jan 9, 2020; and even though the original cause is (more or less) resolved, there have been enough changes in the past six months that the automatic script that blesses new versions still rejects them.

Happily, the un-blessed versions are still made available, just under slightly different URLs.

This is the minimal change needed to use those un-blessed files.

It is theoretically possible that joto, the maintainer of the osmdata box, may do a manual override (and I've requested that here: https://github.com/fossgis/osmdata/issues/12 ) -- but as an alternative, and in the meantime, this enables the coastline data to be updated.

### DIFF
--- a/roles/tile.rb
+++ b/roles/tile.rb
@@ -76,11 +76,11 @@ default_attributes(
     },
     :data => {
       :simplified_land_polygons => {
-        :url => "https://osmdata.openstreetmap.de/download/simplified-land-polygons-complete-3857.zip",
+        :url => "https://osmdata.openstreetmap.de/new/simplified-land-polygons-complete-3857.zip",
         :refresh => true
       },
       :simplified_water_polygons => {
-        :url => "https://osmdata.openstreetmap.de/download/simplified-water-polygons-split-3857.zip",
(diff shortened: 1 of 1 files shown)


tomhughes:
There's a reason for the validation - people are constantly breaking the coastlines.

We are not going to be taking unvalidated coastlines that will flood the world every few days.

Firefishy:
@JesseWeinstein appreciate the effort in creating the PR, but as @tomhughes says, we'd prefer the issue be fixed upstream.

JesseWeinstein:
As I understand it, even the un-blessed ones don't include flooding the world errors; it is just the magnitude of changes cutoff that would be bypassed.

Thanks for considering it, and having this clear statement from such a prominent downstream user may help me motivate upstream to apply a manual reset. :-)

JesseWeinstein:
Osmdata has now made a manual override, so this should be resolved for now, happily. The underlying issue of people controversially marking things as bays remains, though. But now we can deal with it individually.

PR #331, opened 2020-08-14
Align wiki tile layer list with featured tile layers at osm.org

Tigerfell:
I would like to update the list of tile layers of the wiki's MultiMaps extension, so I equals the list of featured tile layers. 

I opened the similar PR #278 in February 2020 including OSM-FR tile layer, but in response to https://github.com/cquest/osmfr-cartocss/issues/65#issuecomment-673613832, I would like to leave that one out for now.


Firefishy:
@gravitystorm Are you ok with adding thunderforest.com cycle style to the OpenStreetMap wiki?

gravitystorm:
> @gravitystorm Are you ok with adding thunderforest.com cycle style to the OpenStreetMap wiki?

Yes, happy to see it.

PR #334, opened 2020-08-20
remove SlippyMap extension from all wikis

Tigerfell:
This is the last one of a series of PRs to replace MediaWiki's SlippyMap extension with MultiMaps extension. The plan and subsequent discussion can be found in #241 (see also Firefishy/SlippyMap/issues/2).

### DIFF
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -467,6 +467,7 @@ action :create do
     repository "https://github.com/Firefishy/SlippyMap.git"
     tag "live"
     update_site false
+    action :delete
   end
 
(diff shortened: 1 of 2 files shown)


Firefishy:
You need to also delete the existing extension, eg: action :delete https://github.com/openstreetmap/chef/blob/master/cookbooks/mediawiki/resources/site.rb#L464

Tigerfell:
I updated the branch and squashed all changes. Does it work now?

Firefishy:
👍  Merged and cleaned up slightly in https://github.com/openstreetmap/chef/commit/2cfa43b04cf27f86e25d4a0b80bd37a5b0ba3e19

PR #350, opened 2020-10-08
Wiki: OpenStreetMap Calendar Extension

thomersch:
The extension enables direct integration of [osmcal](https://osmcal.org) into arbitrary wiki pages through the use of a wiki tag. It doesn't do any injection of code, it just registers the `osmcal` hook and gets called by MediaWiki where necessary. No JavaScript needed.

I tested the code against a bare MediaWiki 1.33 which should match the installed base. As configured, [the code lives in its own repo](https://github.com/thomersch/OSMCALWikiWidget).

Example:

Screenshot 2020-10-08 at 12 36 51

_This is part of [my microgrant](https://wiki.openstreetmap.org/wiki/Microgrants/Microgrants_2020/Proposal/OpenStreetMap_Calendar) work._

### DIFF
--- a/cookbooks/mediawiki/templates/default/mw-ext-osmcalwikiwidget.inc.php.erb
+++ b/cookbooks/mediawiki/templates/default/mw-ext-osmcalwikiwidget.inc.php.erb
@@ -1,3 +1,0 @@
-<?php
-# DO NOT EDIT - This file is being maintained by Chef
-wfLoadExtension( 'OSMCALWikiWidget' );
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -455,7 +455,6 @@ action :create do
 
   mediawiki_extension "OSMCALWikiWidget" do
     site new_resource.site
-    template "mw-ext-osmcalwikiwidget.inc.php.erb"
     repository "https://github.com/thomersch/OSMCALWikiWidget.git"
     tag "live"
     update_site false


cookbooks/mediawiki/templates/default/mw-ext-osmcalwikiwidget.inc.php.erb
```diff
@@ -0,0 +1,3 @@
+<?php
+# DO NOT EDIT - This file is being maintained by Chef
+wfLoadExtension( 'OSMCALWikiWidget' );
```
tomhughes, line 3: There's no need for this if it's just going to call `wfLoadExtension` as we do that by default if a `mediawiki_extension` resource doesn't specify a configuration template.

tomhughes (changes requested):

mmd-osm:
The new extension doesn’t seem to do any local caching. Will that be ok from a load/performance point of view? What happens if osmcal is down?

cookbooks/mediawiki/templates/default/mw-ext-osmcalwikiwidget.inc.php.erb
thomersch: Thanks! I've overlooked that behaviour. Fixed.

thomersch:
> The new extension doesn’t seem to do any local caching. Will that be ok from a load/performance point of view? What happens if osmcal is down?

It does use the MediaWiki's content caching system, though it's unclear to me at which point it is invalidated and reloaded. The API should easily handle a lot of traffic anyway, since it doesn't do any complex lookups.

If osmcal should go down, it would simply show no events:
Screenshot 2020-10-08 at 14 46 21

mmd-osm:
As I mentioned [here](https://wiki.openstreetmap.org/wiki/Talk:Microgrants/Microgrants_2020/Proposal/OpenStreetMap_Calendar) some time ago, the [ExternalData](https://www.mediawiki.org/wiki/Extension:External_Data) has very similar functionality and is supported by mediawiki. They already have some caching in place, in case this turns out to be relevant in any way.

thomersch:
> They already have some own caching mechanism in place, in case this turns out to be relevant in any way.

I'd rather control this via a cache in front of the API instead of having some kind of TTL in some third-party extension.

mmd-osm:
Well, the idea was to get rid of the network roundtrip by using some local caching. I'm not sure if you mean the same thing with "in front of the API".

thomersch:
> Well, the idea was to get rid of the network roundtrip by using some local caching.

As I mentioned in the comment above, it _does_ use MediaWiki's content caching.

mmd-osm:
I have a few more comments:

1. Image link to https://osmcal.org/static/osmcal/favicon.png: do you think you could upload that file to the wiki and use that one instead? Otherwise all wiki users will show up in your server logs and LWG needs to work on the privacy policy topic again.

2. XSS protection: at what point would you deal with a malicious user entering some javascript code in osmcal? Is this stuff being properly escaped before sending it to a browser? I only checked the plugin and didn’t find anything here. Maybe the Python code already handles it?

thomersch:
1. You‘re right, I‘ll throw the image directly into the extension as an asset.

2. Yes, the content is already sanitizied by the form handling, otherwise people could also XSS osmcal itself.

thomersch:
Fixed 1.: https://github.com/thomersch/OSMCALWikiWidget/commit/912f265013ee1582d3b5117b931dbe2bad252255

Firefishy:
Looks good to me.
@thomersch Do you run https://osmcal.org ?

thomersch:
> Looks good to me.
> @thomersch Do you run https://osmcal.org ?

Yep, I do

PR #359, opened 2020-11-21
auto-create *.torrent files (and RSS Feed for them)

mnalis:
This implements:
- Automatic creation of `.torrent` files (by using `mktorrent` program/package) for `pbf/planet-XXXXXX.osm.pbf` and `pbf/full-history/history-XXXXXX.osm.pbf` planet dump files (where `XXXXXX` is date like `201109`).

- RSS/Atom feed for aforementioned `.torrent` files, which allow dedicated torrent clients automatic seeding of new planet dumps. RSS will be available as `/pbf/planet-pbf-rss.xml` and `/pbf/full-history/history-pbf-rss.xml` URIs.

Some notes:
- while I obviously don't have whole OSM infrastructure to test, I did test `mk_rss` and `mk_torrent` bash functions locally, and they produce valid and working `.torrent` and `RSS` files.

- tracker selection: I removed dead trackers discussed in issue #451, and added those that do not have negative affiliations, are announced as public, and were fast and stable for some time. They provide a mix of UDP and HTTP endpoints, and also of IPv4/IPv6. There is also included `http://retracker.local`, sometimes used by endusers and ISPs to automagically provide fast tracker mirror (by simple DNS or `/etc/hosts` change)  

- webseed selection: I've added [all mirrors from wiki](https://wiki.openstreetmap.org/wiki/Planet.osm#Planet.osm_mirrors) which provide at least latest planet in `pbf/planet-XXXXXX.osm.pbf` format (mirrors providing only `planet-latest.osm.pbf` are ignored, as as it might be different file at the moment of torrenting), and support required HTTP range requests (tested). Some of those mirrors (Those listed below `-w https://planet.openstreetmap.org` in `mktorrent` webseed list) do not provide `history-XXXXXX.osm.pbf`, though - it won't affect correct operations, but obviously users won't get as much parallelization there as with `planet-XXXXXX.osm.pbf`. 

Let me know if further work is needed on this PR.

This pull request closes: https://github.com/openstreetmap/operations/issues/451

### DIFF
--- a/cookbooks/planet/templates/default/planetdump.erb
+++ b/cookbooks/planet/templates/default/planetdump.erb
@@ -101,6 +101,56 @@ function mk_torrent {
      -o ${name}.torrent
 }
 
+# Function to create RSS/Atom feed for .torrent files
+function mk_rss {
+  type="$1"		# fixme "history" / "planet"
+  format="$2"		# fixme "pbf"
+  web_dir="$3"		# fixme "pbf/full-history" or "pbf"
+  disk_dir="$4"		# fixme "/store/planet/pbf/full-history"
+  old_pwd="$PWD"
+  rss_name="${type}-${format}-rss.xml"
+  rss_path="${old_pwd}/${rss_name}"
+  rss_baseurl="https://planet.openstreetmap.org"
+  rss_dirurl="${rss_baseurl}/${web_dir}"
+
+  cd "${disk_dir}"
+
+  # RSS header
+  printf '%s\n' \
+     '<?xml version="1.0" encoding="utf-8"?>' \
+     '<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">' \
+     '<channel>' > "${rss_path}"
+  cat >> "${rss_path}" <<__EOF
+     <title>OpenStreetMap planet torrent RSS</title>
+     <link>${rss_baseurl}</link>
+     <atom:link href="${rss_dirurl}/${rss_name}" rel="self" type="application/rss+xml" />
+     <description>RSS feed for ${type}.osm.${format}.torrent</description>
+     <language>en-us</language>
+     <lastBuildDate>`date -R`</lastBuildDate>
+__EOF
+
+  # add RSS item for each torrent
+  for tf in `ls -1t *.torrent | grep -v latest | head -n 5`
+  do
+    cat >> "${rss_path}" <<__EOF
+     <item>
+        <title>$tf</title>
+        <guid>${rss_dirurl}/$tf</guid>
+        <pubDate>`date -R -r $tf`</pubDate>
+        <category>OpenStreetMap</category>
+        <link>${rss_dirurl}/$tf</link>
+        <enclosure url="${rss_dirurl}/$tf" length="`find -maxdepth 1 -name ${tf%.torrent} -printf "%s"`" type="application/x-bittorrent" />
+        <description>OSM Torrent $tf (torrent size: `find -maxdepth 1 -name $tf -printf "%s"`)</description>
+     </item>
+__EOF
+  done
+
+  # RSS footer
+  printf '</channel>\n</rss>\n' >> "${rss_path}"
+  cd "$old_pwd"
+  mv "${rss_path}" "${disk_dir}"
+}
+
 # Function to install a dump in place
 function install_dump {
   type="$1"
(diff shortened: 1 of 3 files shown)


cookbooks/planet/templates/default/planetdump.erb
```diff
@@ -74,6 +74,83 @@ time nice -n 19 /opt/planet-dump-ng/planet-dump-ng \
...
+     -a http://retracker.local/announce \
+     -w https://ftp5.gwdg.de/pub/misc/openstreetmap/planet.openstreetmap.org/${web_path} \
+     -w https://ftpmirror.your.org/pub/openstreetmap/${web_path} \
+     -w https://mirror.init7.net/openstreetmap/${web_path} \
+     -w https://planet.openstreetmap.org/${web_path} \
+     -w https://free.nchc.org.tw/osm.planet/${web_path} \
+     -w https://ftp.fau.de/osm-planet/${web_path} \
+     -w https://ftp.spline.de/pub/openstreetmap/${web_path} \
+     -w https://osm.openarchive.site/${name} \
+     -w https://downloads.opencagedata.com/planet/${name} \
```
pnorman, line 99: Can we add https://ftp.osuosl.org/pub/openstreetmap/?

cookbooks/planet/templates/default/planetdump.erb
```diff
@@ -100,4 +182,8 @@ install_dump "planet" "pbf" "<%= node[:planet][:dump][:pbf_directory] %>"
 install_dump "history" "pbf" "<%= node[:planet][:dump][:pbf_history_directory] %>"
 
 # Remove pbf dumps older than 90 days
-find "<%= node[:planet][:dump][:pbf_directory] %>" "<%= node[:planet][:dump][:pbf_history_directory] %>" -maxdepth 1 -mindepth 1 -type f -mtime +90 \( -iname 'planet-*.pbf' -o -iname 'history-*.pbf' -o -iname 'planet-*.pbf.md5' -o -iname 'history-*.pbf.md5' \) -delete
+find "<%= node[:planet][:dump][:pbf_directory] %>" "<%= node[:planet][:dump][:pbf_history_directory] %>" -maxdepth 1 -mindepth 1 -type f -mtime +90 \( -iname 'planet-*.pbf' -o -iname 'history-*.pbf' -o -iname 'planet-*.pbf.md5' -o -iname 'history-*.pbf.md5' -o -iname 'planet-*.pbf.torrent' -o -iname 'history-*.pbf.torrent' \) -delete
```
pnorman, line 185: Can we break this command onto multiple lines? It's getting excessively long

pnorman (changes requested):
I don't like building XML by string concatenation unless there are unusual circumstances and I have more control of the inputs than here.

cookbooks/planet/templates/default/planetdump.erb
mnalis: ## about XML
I'd agree with handcrafted XML problems if we had more variable input, but the input for generating XML is **quite fixed** here. There are only 3 things that ever change (the rest is extracted to variables for readability and removing redundancy):
  - 6-digit numeric part of name (eg. `planet-201109` becomes `planet-201116`), as created elsewhere in code.
  - purely integer size of file in bytes (as returned by `find -printf "%s"`)
  - date in RFC 5322 format as returned by `date -R` (which ignores locale, and contains only US-ASCII aphanumerics, spaces, `:` for hours/minutes/seconds separation, and `+` or `-` to indicate timezone).

None of the above is problematic for XML generated, nor needs escaping. I've also validated the resulting XML at https://validator.w3.org/feed/ and it is valid. Since input doesn't otherwise change, I think it should be quite good. 
_(But if it is requirement of operations team to separate it in some other programing language which uses proper XML module, I can redo it in perl. But I hope my explanation clears this issue well enough?)_

## about extra webseed
Unfortunately, not at the moment, as mentioned briefly in "webseed selection" section above. https://ftp.osuosl.org/pub/openstreetmap/ provides `.pbf` only with name `planet-latest.osm.pbf` and we need it in format `planet-201109.osm.pbf`. If they can change their system to have latest file in format `planet-YYMMDD.osm.pbf` (and perhaps add `planet-latest.osm.pbf` as symlink if they want it) then they can be added too!
This is important due to race condition when `planet.openstreetmap.org` had published the torrent, but `osuosl.org` didn't yet finished fetching the new planet, then `osuosl.org` would serve blocks from wrong (old) `latest` file, which would poison the swarm with invalid blocks. 
Using `planet-YYMMDD.osm.pbf` as filename fixes that issue, as clients will get `404` if mirror is not yet synced, which they will handle correctly (by trying another peer / webseed)

cookbooks/planet/templates/default/planetdump.erb
mnalis: Agreed. Now split in https://github.com/openstreetmap/chef/pull/359/commits/345e1df41a35cd7143f21fe7a2ba5f58800a93ba

Firefishy:
Merged via #360 
I have removed the RSS functionality for the moment.

PR #362, opened 2020-11-25
Add croatian mirror to torrent webseed

hbogner:
Add Croatian mirror as a webseed for .pbf


Firefishy:
I'd prefer not to add a mirror that only features PBF files. That code path is used for all types of files.

hbogner:
I plan to mirror everything that will be torented, just wasn't sure what that includes.
For .bz2 I have to add it in /planet/year, like on https://planet.openstreetmap.org/planet/2020/
Fixed it and started bz2 on the mirror

PR #368, opened 2020-12-03
Ensure lockfile directory for changeset replication exists.

zerebubuth:
fcf52fb06a5ddb3e1cbb54164511aae2b135a80e altered the [lockfile directory for changeset replication](https://github.com/openstreetmap/chef/commit/fcf52fb06a5ddb3e1cbb54164511aae2b135a80e#diff-045e8c1decbd8a4f8465aa40ca33722f2c1c69fc629d057fd434be318f79a966R4), but that directory wasn't created in the recipe.


tomhughes:
No the problem is just that I typoed the path.

PR #376, opened 2020-12-27
establish confirmed user's group in public wikis

Tigerfell:
closes https://github.com/openstreetmap/operations/issues/488

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -229,6 +229,13 @@ $wgGroupPermissions['*']['createaccount'] = false;
 $wgUploadPath = "$wgScriptPath/img_auth.php";
 <% end -%>
 
+<% if not(@mediawiki[:private_accounts]) and not(@mediawiki[:private_site]) -%>
+# user group "confirmed" with identical rights as "autoconfirmed", but assigned manually by sysops
+$wgGroupPermissions['confirmed'] = $wgGroupPermissions['autoconfirmed'];
+$wgAddGroups['sysop'][] = 'confirmed';
+$wgRemoveGroups['sysop'][] = 'confirmed';
+<% end -%>
+
 # Allow Subpages on Main Namespace
 $wgNamespacesWithSubpages[NS_MAIN] = true;
 


Firefishy:
Could you explain the groups a bit more, I don't understand the not and what a user's groups end up as.

Tigerfell:
`<% if not(@mediawiki[:private_accounts]) and not(@mediawiki[:private_site]) -%>`
You wanted to have the change limited to wiki.osm.org. There is no separate configuration file for wiki.osm.org (#257 would create one when it works as intended). I interpreted that as "all wikis that neither are private nor use private accounts". 
`$wgGroupPermissions['confirmed'] = $wgGroupPermissions['autoconfirmed'];`
That line copies all permissions from the autoconfirmed user group to a user group called "confirmed". New user groups are created by assigning permissions to a not yet used group name ([docs](https://www.mediawiki.org/wiki/Manual:User_rights#Creating_a_new_group_and_assigning_permissions_to_it)). If one later decides to change the permissions for autoconfirmed users a few lines above the ones to be changed, the permissions of confirmed users stay aligned. I was not sure if one can actually copy arrays in PHP this way, so I successfully tested it in a local MediaWiki installation and looked at [Wikimedia's meta wiki configuration](https://phabricator.wikimedia.org/source/mediawiki-config/browse/master/wmf-config/CommonSettings.php;f69daf462e76cc75373641f91ecf602d1dfdcf8e$3892) which happens to use the exact same line. 
`$wgAddGroups['sysop'][] = 'confirmed';`
This line grants the sysops a.k.a. wiki administrators the ability to assign this one user group to everyone as proposed in https://github.com/openstreetmap/operations/issues/488#issue-773130252. Bureaucrats do not need this, they have the [right to assign and un-assign all user groups](https://www.mediawiki.org/wiki/Manual:$wgAddGroups) to all users. 
`$wgRemoveGroups['sysop'][] = 'confirmed';`
I would grant Sysops the right to remove the confirmed status, too. I was thinking of a situation where a sysop disagrees about the decision of an other sysop or a confirmed user abuses the limited permissions while the granting sysop is offline. That is mostly hypothetical, because the user can simply be blocked which is more effective and does not depend on this user group at all. 

Alternatively, one could change the configuration of the autoconfirmed user group to  have one user group that will be granted either automatically or manually. I decided against it, because it will effect more users and requires more changes: The current autoconfirm mechanism apparently "[does not actually add users to a group](https://www.mediawiki.org/wiki/Manual:$wgAutopromote#Caveats)" and it does not allow the assignment or revocation of the autoconfirmed user group (except for abuse filters). Consequently, the promotion process would need to be reconfigured, too, and there were two promotion processes (one for wiki.osm.org, one for the rest). The name of the group would probably need to be changed to avoid confusion. Additionally, all wikis of Wikimedia seem to use the configuration proposed by me, so it is proven to work. Lastly, I expect that all "confirmed" users will eventually receive promotion into the "autoconfirmed" user group. A revocation of the "confirmed" status does not have any effect for autoconfirmed users, because the "autoconfirmed" user group grants the same rights. That effectively limits the power of revocation. In contrast, the alternative configuration change would allow sysops to un-assign even established users from the "confirmed"/"autoconfirmed" user group.

Firefishy:
Looks good. I will merge when I get a chance.

Tigerfell:
@Firefishy ?

matkoniecz:
Is it OK to ping again? One more user run into this issue.

PR #377, opened 2020-12-28
Redirect fix for HSTS preloading of osm.org domain

Firefishy:


Firefishy:
This is to fix the redirect to allow preloading of osm.org domain (not to be confused with openstreetmap.org domain): https://hstspreload.org/?domain=osm.org

pnorman (approved):

PR #378, opened 2020-12-30
Only reserve one CPU for other tasks on render servers

pnorman:
I had a look at the CPU usage of our render servers. Below is Odin, which is typical

image

They never reach 100% utilization, which is good, but I think we're leaving too much idle right now.

I graphed the idle CPU with the PromQL `sum by (instance) (rate(node_cpu_seconds_total{instance=~"albi|odin|ysera|rhaegal",mode="idle"}[$__rate_interval]))` and looked at a peak period where those four servers were capped in CPU and found they all had at least 2 threads idle.

pyrene and bowser did not hit a CPU cap.


tas50:
Any thought on providing the service with all possible threads to use and instead using the various systemd CPU controls to set the priority on the unit lower? That would potentially let you run the servers at 100% without impacting other processes.

PR #400, opened 2021-03-09
Create "Proposal" namespace for main wiki 2

Tigerfell:
I fixed the template linking issue in #257 and opened a new PR based on previous work. The old description follows below the line.
________

This PR configures a new namespace "Proposal" for the main wiki. It will have a slightly lower search rank as the tag/key pages. Additionally, it moves the namespace config to a new file in the wiki cookbook. This will remove the language namespaces for all other wikis.
Before deploying please check

- [ ] none of the other wikis should have any pages in the language namespaces like "FR" or "FR talk". You can use [Special:PrefixIndex](https://wiki.openstreetmap.org/wiki/Special:PrefixIndex).
- [ ] that the job queues of the wikis are empty as explained in https://www.mediawiki.org/wiki/Manual:Using_custom_namespaces
- [ ] temporarily remove the two pages starting with the same prefix "Proposal:" in the main wiki. You can restore them after the config change.

I created four commits, so you can squash them together as you like. 

Related discussion https://wiki.openstreetmap.org/wiki/Talk:Wiki/Archive_6#Proposal:_new_namespace_for_tagging_proposals and again starting with https://lists.openstreetmap.org/pipermail/talk/2019-April/082453.html. I asked in mid-August if someone would object when I would open this PR and nothing happened.


Firefishy:
Merged and the changes done.

The 3 parts of this would have been easier if split in 3 parts: 
1) Adding Proposal Namespace
2) Modifying the Search
3) Refactor to use includes (removed as they did not work)

PR #411, opened 2021-05-12
Remove support of doc files

Tigerfell:


tomhughes:
What's the context? Why do we want to remove support for them? I imagine somebody asked for it at some point...

Please don't open PRs with absolutely no explanation of why you are proposing a change!

Firefishy:
Wikis allowing the upload of Word documents are somewhat rare. It was a mistake me enabling it in the first place.

PR #431, opened 2021-07-04
Run tilelog earlier in the day

pnorman:
The logs are pushed to S3 within minutes, so this can run much earlier in the day.

Depending on when this PR is merged, a manual run might be required.

### DIFF
--- a/cookbooks/tilelog/recipes/default.rb
+++ b/cookbooks/tilelog/recipes/default.rb
@@ -50,8 +50,8 @@ template "/usr/local/bin/tilelog" do
 end
 
 cron_d "tilelog" do
-  minute "17"
-  hour "22"
+  minute "7"
+  hour "1"
   user "www-data"
   command "/usr/local/bin/tilelog"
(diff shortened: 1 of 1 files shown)


tomhughes:
Within minutes, but in what timezone? That's why it ran when it did I think to allow time for all the machines to send in that day's logs.

Obviously in the new system that will be different - are you doing everything on the basis of UTC times? If so then be aware that this machines is running on Amsterdam time for cron purposes, so currently UTC+2.

pnorman:
If it's in UTC+2 I'll adjust the time.

With the new system log files are sent every 10 minutes starting on the hour, so it'll only ever be slightly more than 10 minute behind.

tomhughes:
If we just want to run it at UTC+n then probably best is to convert it to a systemd timer rather than relying on knowing what timezone the server is in - the whole thing about cron running in local time was mostly to help with the tile caches to let us run things "out of hours" across all the caches. Other than that the machines run in UTC>

PR #440, opened 2021-08-11
Add docker support to shorewall

Firefishy:
Enable docker support in shorewall.

Currently shorewall will break docker's networking on a restart


tomhughes (changes requested):
I've done some playing and I think we can resolve my concerns about changing `clear` to `stop` by making one additional change to add `/etc/shorewall/stoppedrules` with the following content:

```
ACCEPT	-	-
ACCEPT	-	$FW
```

That means that it will accept all traffic in the stopped state, as it does now when the service is stopped.

tomhughes (approved):
Looks good to me.

PR #442, opened 2021-08-14
Increase usability of anti-spam features

Tigerfell:
The first commit enables logging for two anti-spam wiki extensions. Currently, both [Spam blacklist log](https://wiki.openstreetmap.org/wiki/Special:Log?type=spamblacklist) and [Title blacklist log](https://wiki.openstreetmap.org/wiki/Special:Log?type=titleblacklist) are empty because logging is not enabled. I would like to use the logs to find spammers as well as wrong hits. By default, one log is visible to all signed-in users the other one is visible to administrators only. I explicitly set the viewing permissions to avoid changes with updates. 

The second commit removes the spam regex from the configuration files. I created a [similar abuse filter](https://wiki.openstreetmap.org/wiki/Special:AbuseFilter/11) instead. Abuse filters are editable by wiki administrators. The regex can be combined with other conditions such as the account age.

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -242,9 +242,6 @@ $wgEmailConfirmToEdit = true;
 # Extend autoblock period
 $wgAutoblockExpiry = 7776000; // 90 days
 
-# Spam filter regex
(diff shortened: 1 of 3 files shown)


Firefishy (approved):

Tigerfell:
The two failing test are coverage errors in Prometheus and SotM webpage. I do not see a relation to this PR.

PR #447, opened 2021-08-28
update civicrm to 5.40.4

datendelphin:


tomhughes:
You needed to run the database migration after I did the upgrade! I've run it now (to try and resolve https://github.com/openstreetmap/operations/issues/566) but it warned about some template that might need fixing:

image

datendelphin:
Thank you for the info. Didn't know I had to call the database migration manually, should have read the upgrade guide.

PR #462, opened 2021-11-07
remove account confirmation for private wikis

Tigerfell:
### Explanation
The pull request removes the automatic account confirmation mechanism for "private account" wikis and "private site" wikis. It first sets the required number of edits and the required account age to 0 and then merges and deletes the "autoconfirmed" user group (identical as how the "interface admin" user group is deleted). The rest of the change moves code in the configuration file and adds comments. This is done to improve readability and to load extensions before manipulating those user rights. [In a separate file](https://github.com/openstreetmap/chef/blob/6c869d1b5625e8cbe1c091ed869c162bc9a6db59/cookbooks/mediawiki/templates/default/mw-ext-ConfirmEdit.inc.php.erb), the ConfirmEdit extension includes the configuration ```$wgGroupPermissions['autoconfirmed']['skipcaptcha'] = true;``` that is to be adapted to "all users skip the CAPTCHA" in case of private account and private site wikis or "autoconfirmed and manually confirmed users skip the CAPTCHA" in case of public wikis. This applies similarly to TimedMediaHandler extension. 

### Reasoning
Account confirmation is one measure against spam. When new accounts are created, one usually does not know if they are trustworthy. As a result, some hard-to-fix actions are disabled for them. This approach is reasonable for wikis that allow signing up for new accounts. In private site wikis, administrators are allowed to create new accounts only. In private account wikis, existing users are allowed to create new accounts only. In both cases, new users are trustworthy. As @tordanik explained, the present system of account confirmation even hinders onboarding of new board members because the user rights can not be granted manually. As [discussed previously](https://github.com/openstreetmap/operations/issues/488#issuecomment-749725192) account confirmation for non wiki.osm.org should be automatic. As a result, one ends up with this solution. 

### Request/Issue
The pull request fixes openstreetmap/operations#517.

### DIFF
--- a/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
+++ b/cookbooks/mediawiki/templates/default/LocalSettings.php.erb
@@ -185,16 +185,6 @@ $wgResourceLoaderMaxQueryLength = -1;
 # Only Allow Signed-in users to edit
 $wgGroupPermissions['*']['edit'] = false;
 
-# Only allow autoconfirmed for a few actions
-$wgGroupPermissions['user']['move'] = false;
-$wgGroupPermissions['user']['movefile'] = false;
-$wgGroupPermissions['user']['move-categorypages'] = false;
-$wgGroupPermissions['user']['upload'] = false;
-$wgGroupPermissions['autoconfirmed']['move'] = true;
-$wgGroupPermissions['autoconfirmed']['movefile'] = true;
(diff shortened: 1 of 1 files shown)


Firefishy (approved):

Firefishy:
@Tigerfell This is a really appreciated improvement. I am sorry that it has taken me such a crazy length of time to review it and get it merged. 👍

PR #471, opened 2021-12-05
imagery: Add OS OpenMap Local Oct 2021

Firefishy:
Add OS OpenMap Local Oct 2021.

Do not yet merge, kessie is still building the gdaladdo. ETA 8 Dec 2021.


Firefishy:
Want to watch? Here you go: https://munin.openstreetmap.org/openstreetmap.org/kessie.openstreetmap.org/cpu.html

Firefishy:
Latest gdaladdo running multiple threads is a LOT faster.

PR #480, opened 2022-02-13
Enable BBR at Dublin with sufficient uplink and switch capacity

Firefishy:
Equinix Dublin has a 10Gb uplink and significantly faster switches.

We had to disable BBR in Equinix Amsterdam due to the uplink going over capacity causing switch/router port buffer overrun and packet loss.

The same scenario is very unlikely to happen in Dublin with machines limited to 2x 1Gb each and uplink allowing 10Gb.

### DIFF
--- a/roles/equinix-dub.rb
+++ b/roles/equinix-dub.rb
@@ -2,6 +2,15 @@ name "equinix-dub"
 description "Role applied to all servers at Equinix Dublin"
 
 default_attributes(
+  :sysctl => {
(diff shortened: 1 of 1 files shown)


Firefishy:
@tomhughes @pnorman Any view on this?

tomhughes:
Not really. I mean the only way to know for sure is to try it and even then we'll only know when there is significant load.

Firefishy:
Let us give it a try, we can always revert if any issues.

PR #483, opened 2022-02-25
Extend chef discourse/community management

Firefishy:
First stab at complete management of discourse/community using chef.


Firefishy:
Still to be completed is adding backups.

cookbooks/community/templates/default/web_only.yml.erb
```diff
@@ -0,0 +1,131 @@
...
+
+  ## TODO: List of comma delimited emails that will be made admin and developer
+  ## on initial signup example 'user1@example.com,user2@example.com'
+  DISCOURSE_DEVELOPER_EMAILS: 'operations@openstreetmap.org'
+
+  ## TODO: The SMTP mail server used to validate new accounts and send notifications
+  # SMTP ADDRESS, username, and password are required
+  # WARNING the char '#' in SMTP password can cause problems!
+  DISCOURSE_SMTP_ADDRESS: mail.openstreetmap.org
+  DISCOURSE_SMTP_PORT: 26
```
tomhughes, line 51: Does discourse have internal queuing with retries? If not then I would normally suggest using localhost:25 as the server so that mail can be queued locally in the event of problems reaching the main server.

cookbooks/community/templates/default/web_only.yml.erb
```diff
@@ -0,0 +1,131 @@
...
+  DISCOURSE_SMTP_USER_NAME:
+  DISCOURSE_SMTP_PASSWORD:
+  # DISCOURSE_SMTP_ENABLE_START_TLS: true           # (optional, default true)
+  DISCOURSE_SMTP_DOMAIN: community.openstreetmap.org
+  DISCOURSE_NOTIFICATION_EMAIL: noreply@community.openstreetmap.org
```
tomhughes, line 56: Probably better to use `community@noreply.openstreetmap.org` here and then drop a file in `cookbooks/exim/files/default/noreply/community` which will act as an autoresponse to any mails.

cookbooks/community/templates/default/web_only.yml.erb
```diff
@@ -0,0 +1,131 @@
...
+
+  ## TODO: The SMTP mail server used to validate new accounts and send notifications
+  # SMTP ADDRESS, username, and password are required
+  # WARNING the char '#' in SMTP password can cause problems!
+  DISCOURSE_SMTP_ADDRESS: mail.openstreetmap.org
+  DISCOURSE_SMTP_PORT: 26
+  DISCOURSE_SMTP_USER_NAME:
+  DISCOURSE_SMTP_PASSWORD:
+  # DISCOURSE_SMTP_ENABLE_START_TLS: true           # (optional, default true)
+  DISCOURSE_SMTP_DOMAIN: community.openstreetmap.org
```
tomhughes, line 55: This is so that mail can be routed back to discourse? Which presumably means we need a new route adding in `roles/mail.rb` and also something in the discourse cookbook to have exim route mail into discourse.

cookbooks/community/templates/default/web_only.yml.erb
Firefishy: Yes, it used an internal sidekiq queue. https://github.com/discourse/discourse/blob/main/app/jobs/regular/user_email.rb#L5-L28

cookbooks/community/templates/default/web_only.yml.erb
Firefishy: Added to PR.

cookbooks/community/templates/default/web_only.yml.erb
Firefishy: Added to PR

cookbooks/exim/files/default/noreply/community
```diff
@@ -0,0 +1,10 @@
+This is an automated response to your email, which was sent to an
+unattended address.
+
+If you were trying to reply to a question, answer or comment on
+the community.openstreetmap.org web site then you should do so using
+the web site as it is not possible to reply by email.
```
tomhughes, line 6: Is this really true? Are we not going to support mailing list mode? Is that not what the mail receiver stuff is for?

roles/mail.rb
```diff
@@ -46,6 +46,11 @@
         :comment => "join.osmfoundation.org",
         :domains => ["join.osmfoundation.org"],
         :host => "ridley.ucl.openstreetmap.org"
+      },
+      :discourse => {
+        :comment => "community.osmfoundation.org",
+        :domains => ["community.osmfoundation.org"],
+        :host => "lockheed.ams.openstreetmap.org"
```
tomhughes, line 53: That will get the email to lockheed, but how does it then get into discourse? The exim on lockheed will somehow need to route it to the discourse mail receiver container?

cookbooks/exim/files/default/noreply/community
Firefishy: Not 100% true. It depends on the type of email sent. Digest and many Notification emails cannot be replied to.
The messages that can be replied to will be use another from email.

roles/mail.rb
Firefishy: The `mail-receiver` container runs postfix with discourse handler. [Standard config it runs on port 25](https://github.com/openstreetmap/chef/pull/483/files#diff-4b7f61680f19b3fe1dded83774a2388ebc80e0bdb661f92a1822a3f4490ce1f0R14), but these can be mapped to another port if exim would be able to handle that.

tomhughes: Well it won't be able to bind port 25 because exim will already be there, unless you do something to either stop exim running or stop it listening on port 25.

cookbooks/exim/files/default/noreply/community
tomhughes: Right but normal replies to non-digest questions and answers should work so maybe the wording could be better? Guess it would need to be fairly vague though, just "you can't reply to this email so please use the web site" or something?

Firefishy: Removed the text a bit. I will improve it more once I see the service running a bit.

roles/mail.rb
Firefishy: Added ::2500 and moved the docker external port to 2500 for the `mail-receiver`

tomhughes (approved):
I have no idea about most of this but the bits I do know about look OK now.

PR #489, opened 2022-03-31
openssh: disable password authentication by default (except dev)

Firefishy:
Disable openssh password authentication by default.

Closes: https://github.com/openstreetmap/operations/issues/603

### DIFF
--- a/cookbooks/openssh/attributes/default.rb
+++ b/cookbooks/openssh/attributes/default.rb
@@ -1,1 +1,2 @@
 default[:openssh][:port] = 22
+default[:openssh][:password_authentication] = false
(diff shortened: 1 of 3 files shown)


cookbooks/openssh/attributes/default.rb
```diff
@@ -1 +1,2 @@
 default[:openssh][:port] = 22
+default[:openssh][:disable_password_authentication] = true
```
tomhughes, line 2: I would reverse the sense of this and just call it `:password_authentication` which is both short and matches the actual ssh configuration option.

tomhughes (approved):

PR #491, opened 2022-04-03
Revert "mediawiki: add QuickInstantCommons extension"

Firefishy:
Reverts openstreetmap/chef#490

### DIFF
--- a/cookbooks/mediawiki/resources/site.rb
+++ b/cookbooks/mediawiki/resources/site.rb
@@ -503,12 +503,6 @@ action :create do
     update_site false
   end
 
-  mediawiki_extension "QuickInstantCommons" do
-    site new_resource.site
-    update_site false
-    only_if { new_resource.commons == true }
-  end
-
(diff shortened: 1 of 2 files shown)


Firefishy:
Code throws an exception. Doesn't feel like this is ready to prime time to me
```
[Sun Apr 03 02:01:47.558278 2022] [proxy_fcgi:error] [pid 2068696:tid 140417622329088] [client 2a01:4f9:6b:1421::2:42500] AH01071: Got error 'PHP message: PHP Warning:  Could not check compatibility between MediaWiki\\Extension\\QuickInstantCommons\\File::getDescription($audience = self::FOR_PUBLIC, ?MediaWiki\\Permissions\\Authority $performer = NULL) and File::getDescription($audience = self::FOR_PUBLIC, ?User $user = NULL), because class MediaWiki\\Permissions\\Authority is not available in /srv/wiki.openstreetmap.org/w/extensions/QuickInstantCommons/src/File.php on line 445'
```

PR #492, opened 2022-04-13
github actions: Switch to action ruby/setup-ruby

Firefishy:
Switch action from [ruby/setup-ruby](https://github.com/ruby/setup-ruby) instead of deprecated [actions/setup-ruby](https://github.com/actions/setup-ruby)


.github/workflows/cookstyle.yml
```diff
@@ -10,18 +10,8 @@ jobs:
...
-        path: vendor/bundle
-        key: bundle-${{ runner.OS }}-${{ hashFiles('Gemfile.lock') }}
-        restore-keys: |
-          bundle-${{ runner.OS }}-
-    - name: Install gems
-      run: |
-        gem install bundler
-        bundle config set deployment true
-        bundle install --jobs 4 --retry 3
+        bundler-cache: true
```
tomhughes, line 15: I know this is what they tell you to do but I'm not convinced it correctly encodes the OS version... The closest I can see is that I think it uses `common.getVirtualEnvironmentName()` as part of the cache key.

The use of `runner.OS` in our keys was very deliberate because we had issues when `ubuntu-latest` changed from one version to the next with needing to make sure the cache was invalidated.

.github/workflows/test-kitchen.yml
```diff
@@ -106,22 +106,12 @@ jobs:
...
-          bundle-${{ runner.OS }}-
+        bundler-cache: true
     - name: Disable apparmor for mysqld
       run: |
         sudo apt-get install apparmor-utils
         sudo aa-disable /usr/sbin/mysqld
-    - name: Install gems
-      run: |
-        gem install bundler
-        bundle config set deployment true
```
tomhughes, line 124: Do we need to preserve this? There are details for how to replace `bundle config` in the `setup-ruby` doco but it might not matter in this case as I think this might be to do with our caching?

.github/workflows/cookstyle.yml
tomhughes: Actually I think that does work, given the code at https://github.com/ruby/setup-ruby/blob/master/common.js#L104.

.github/workflows/test-kitchen.yml
tomhughes: Look like this was a translation from using `--deployment` with bundler 1.x and that in turn was probably just copied from the example so this is probably fine to go as well.

PR #493, opened 2022-04-14
stateofthemap: Use global bundle caching

Firefishy:
Use global caching of gem bundles, which should speed up tests a lot

### DIFF
--- a/cookbooks/stateofthemap/attributes/default.rb
+++ b/cookbooks/stateofthemap/attributes/default.rb
@@ -1,0 +1,1 @@
+default[:accounts][:users][:stateofthemap][:status] = :role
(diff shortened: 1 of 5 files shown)


tomhughes:
Is that repo using bundler 2 though? I suspect not given it is still using the `--deployment` switch which I believe is removed in bundler 2.

Firefishy:
Looks like no. I will upgrade the sotm repos shortly.

tomhughes:
Once you do that you will need to merge this - once the repo uses bundler 2 chef will to as well.

tomhughes:
I'd like to understand more about what the option does though - presumably it's not really global because permissions would make that impossible so presumably it is per-user? Is there documentation somewhere?

tomhughes:
Oh that bundle is being installed as root anyway.

tomhughes:
Just give up, it's clearly not worth it if your global cache is now not global at all and you've had to abandon our standardised structure of `/srv` to get there!

Firefishy:
Giving up (for now)... There is some weird gem / bundler action happening.

PR #502, opened 2022-06-25
Add new SSH key for Matt

zerebubuth:
Since #489, I should be using an SSH key to log in, but it looks like I didn't have one set up (oops).


tomhughes:
Merged, but I have already fixed this mornings issue with the changeset feed by manually renaming the temporary state file over the main state file which was empty.

zerebubuth:
Yes, I noticed the failure emails stopped. Thanks for that!

I was logging in to fix it when I noticed that I couldn't any more. :facepalm:

Firefishy:
Sorry! I should have checked more extensively that everyone had good SSH keys.

zerebubuth:
Haha, no worries. It's pretty indicative of how often I need to log in that it was almost 3 months before I made this PR :sweat_smile: (Actually, longer, since I didn't actually have to log in - Tom already fixed the problem I was getting alerts for!)

PR #509, opened 2022-07-10
Remove double slash in Vespucci IRC links

pnorman:


pnorman:
CI failures seem unrelated and caused by a connection problem.

PR #514, opened 2022-07-14
Block requests for tiles unless they come from a CDN node

tomhughes:
A first attempt at implementing https://github.com/openstreetmap/operations/issues/679.

Things it still needs:

* Allowing statuscake check ranges (do we have a source for these?)
* Allowing OSMF IP ranges (this would be a massive pain though we could do the main data centres easily)

### DIFF
--- a/cookbooks/tile/recipes/default.rb
+++ b/cookbooks/tile/recipes/default.rb
@@ -29,6 +29,7 @@ include_recipe "ruby"
 include_recipe "tools"
 
 blocks = data_bag_item("tile", "blocks")
+admins = data_bag_item("apache", "admins")
 web_passwords = data_bag_item("web", "passwords")
 
 apache_module "alias"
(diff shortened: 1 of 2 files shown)


pnorman:
> * Allowing OSMF IP ranges (this would be a massive pain though we could do the main data centres easily)

I think Dublin and Amsterdam is enough.

tomhughes:
I've added the statuscake IPs now.

grischard:
What will be the error message for unauthorised IP addresses, 403?

tomhughes:
I think so yes.

Firefishy:
@pnorman @tomhughes Is this ready to go? Should we communicate anything out about this change?

tomhughes:
I believe it is ready, yes.

pnorman (approved):

pnorman:
I don't see any need for announcements. No one should be using our internal addresses

tomhughes:
Sadly this doesn't work because the client IP has already been mapped to the real IP by the time the check is done. I was hoping that putting the IP check before we configured the mod_remoteip stuff would avoid that but it doesn't :-(

PR #515, opened 2022-07-19
hardware: Set OSM ilo defaults

Firefishy:
- Automatically power on servers after random interval
- HP Dynamic Power Savings mode (HP Default)
- Set server name in ilo
- Unset max power (Watt) cap.

Link: https://github.com/openstreetmap/operations/issues/686

### DIFF
--- a/cookbooks/hardware/templates/default/ilo-defaults.xml.erb
+++ b/cookbooks/hardware/templates/default/ilo-defaults.xml.erb
@@ -1,0 +1,21 @@
+<?xml version="1.0"?>
+<RIBCL VERSION="2.0">
+  <LOGIN USER_LOGIN="admin" PASSWORD="password">
+    <SERVER_INFO mode="write">
+      <!-- Set Server Name -->
+      <SERVER_NAME VALUE="<%= node[:fqdn] %>"/>
+      <!-- 1 Operating system control mode -->
+      <!-- 2 HP Static Low Power mode -->
+      <!-- 3 HP Dynamic Power Savings mode -->
+      <!-- 4 HP Static High Performance mode -->
+      <SET_HOST_POWER_SAVER HOST_POWER_SAVER="3"/>
+      <!-- Set Auto Power ON -->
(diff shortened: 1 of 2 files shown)


Firefishy:
I've used the path `/etc/ilo-defaults.xml` for the config, but the file can really be anywhere.

Firefishy:
The "username" and "password" are required, but ignored for local config.

cookbooks/hardware/templates/default/ilo-defaults.xml.erb
```diff
@@ -0,0 +1,21 @@
...
+      <SERVER_NAME VALUE="<%= node[:fqdn] %>"/>
+      <!-- 1 Operating system control mode -->
+      <!-- 2 HP Static Low Power mode -->
+      <!-- 3 HP Dynamic Power Savings mode -->
+      <!-- 4 HP Static High Performance mode -->
+      <SET_HOST_POWER_SAVER HOST_POWER_SAVER="3"/>
+      <!-- Set Auto Power ON -->
+      <!-- Set Auto Power after random delay -->
+      <!-- Both are required -->
+      <SERVER_AUTO_PWR VALUE="ON"/>
```
tomhughes, line 15: What exactly does this mean? Presumably it won't power back on if you power it down so is this just after a power restoration? What if the machine was powered off before the power loss?

Firefishy: On power restore automatically power on the system. Opens are ON, OFF, LAST.

Firefishy: > What if the machine was powered off before the power loss?
I think the system will then power ON again after the power loss. I have tended to always set to ON.

tomhughes: I usually go for LAST on PCs but it's not reliable then fine - it's true our machines do normally mysteriously power up after a power loss even if they were off before.

PR #517, opened 2022-07-25
Update to tilelog v1.4.0 and generate parquet files

pnorman:
- performance at generating the tile list file is substantially improved
- GZIP CSV logs are converted to parquet before files are made, resulting in less storage used and faster queries

As the queries now look at successful tile requests instead of all successful requests, usage numbers are not identical, but the changes are very minor.


pnorman:
Note: If tilelog is run twice in a row on the same day, it will error when it tries to populate the parquet tables twice with the same data.

PR #518, opened 2022-07-28
networking: Add grant roaming wg

Firefishy:
Add my roaming wireguard peer.


tomhughes:
What's the logic behind adding this rather than just removing the endpoint on the other entry - are you expecting to have them both connected at the same time? If so then might it be better to use a different key pair?

Firefishy:
> What's the logic behind adding this rather than just removing the endpoint on the other entry - are you expecting to have them both connected at the same time? If so then might it be better to use a different key pair?

Damn it. I forgot to update the key. Both will be connected at the same time.

tomhughes (approved):

PR #519, opened 2022-07-29
Add WebGL tracker plugin to piwik

pnorman:
I'm not sure how to test this change - is there any docs on how to run the piwik cookbook locally?


tomhughes:
I was going to look at this today but I don't think that's anything like enough - something will need to actually checkout the code to start with!

pnorman:
I don't see anything that checks out the code for the other plugins (e.g. DevicesDetection). My guess is that it gets it from https://plugins.matomo.org/DeviceFeatureWebGL

pnorman:
Reading the docs there might be other steps - but I can't figure out how they were done for other plugins.

tomhughes:
We've never used an external plugin before - all the current ones are part of the default install.

PR #521, opened 2022-08-02
wiki: switch to using hcaptcha

Firefishy:
Fixes: https://github.com/openstreetmap/operations/issues/454


Firefishy:
hcaptcha login credentials are in the Ops gopass store.

PR #261, opened 2020-01-25
Change Wikibase site name to Data Items

Tigerfell:
By default, it has the same name as the wiki instance which is misleading in case of "OpenStreetMap wiki", implements openstreetmap/operations#352

Documentation is located at https://phabricator.wikimedia.org/diffusion/EWBA/browse/master/docs/options.wiki;ad3f41c6e3692dc1e897ea6e70e08451d06a59b3$238.

PR #523, opened 2022-08-03
Change Wikibase site name to Data Items

matkoniecz:
By default, it has the same name as the wiki instance which is misleading in case of "OpenStreetMap wiki", implements openstreetmap/operations#352

Resurrects #261 which was not processed for over a year and finally closed by the author.

Fixes https://github.com/openstreetmap/operations/issues/352

While I am strongly disliking how data items turned out, I can confirm that for naming them data items is a clear consensus.

### DIFF
--- a/cookbooks/wiki/templates/default/mw-ext-Wikibase.inc.php.erb
+++ b/cookbooks/wiki/templates/default/mw-ext-Wikibase.inc.php.erb
@@ -65,6 +65,7 @@ call_user_func( function() {
 // $wgWBClientSettings['showExternalRecentChanges'] = true;
 
 $wgWBClientSettings['namespaces'] = [ NS_MAIN ];
+$wgWBClientSettings['repoSiteName'] = "Data Items";
 
 // Avoid complaints that nobody seems to know the cause off...
 $wgWBClientSettings['entityUsagePerPageLimit'] = 500;


matkoniecz:
@Tigerfell Just to confirm it - you closed it because it was not processed for over year, right? Not because it stopped working or for some other reasons?

matkoniecz:
wrong order, I should asked @Tigerfell first, sorry

tomhughes:
If this does get reopened how sure are we that it isn't going to break things? Wikibase is incredibly fragile so we need to be very careful...

cookbooks/wiki/templates/default/mw-ext-Wikibase.inc.php.erb
```diff
@@ -65,6 +65,7 @@ call_user_func( function() {
 // $wgWBClientSettings['showExternalRecentChanges'] = true;
 
 $wgWBClientSettings['namespaces'] = [ NS_MAIN ];
+$wgWBClientSettings['repoSiteName'] = "Data Items";
```
1ec5, line 68: The [`repoSiteName`](https://doc.wikimedia.org/Wikibase/master/php/md_docs_topics_options.html#autotoc_md357) variable should be set to a message key. That way, administrators can translate the site name into other languages on-wiki using MediaWiki: messages.

PR #528, opened 2022-08-04
Block requests for tiles unless they come from a CDN node

tomhughes:
Attempt number two at restricting direct access to the render servers...

This one has an added rule that allows requests where the remapped IP address from mod_remoteip is different to the real address of the connection.

It has been tested briefly on ysera and appears to work.


grischard:
Do we want to send back a 403, or a 302 to a blocked.png that tells people to use tile.openstreetmap.org?

pnorman:
Is this ready to go?

tomhughes:
As far as I know, yes.

PR #529, opened 2022-08-08
Fix $wgExtraNamespaces[NS_PROPOSAL_TALK]

not-my-profile:
As per https://www.mediawiki.org/wiki/Manual:Using_custom_namespaces
the string values in $wgExtraNamespaces have to use underscores
instead of spaces ... otherwise MediaWiki wrongly recognizes
"Proposal_talk:Foo" to be in the main namespace.


Firefishy:
Thank you.

PR #532, opened 2022-08-13
Name php-fpm socket to work with apparmor-profiles

Firefishy:
`apparmor-profiles` package on errol (and elsewhere) provides a default php-fpm profile.
The socket is expected with name: `@{run}/php{,-fpm}/php*-fpm.sock`


Firefishy:
`apparmor-profiles` runs in complain only mode (not enforced). I am unsure under what scenario it became installed on `errol` (others too?) but the profile might be helpful for noticing bad actor activity.

tomhughes (approved):
Assuming you've found all the uses it looks fine to me.

PR #533, opened 2022-08-26
Renice background services

Firefishy:
Renice background service to allow full speed important services.
Reniced services will be ignored for ondemand CPU scaling saving power.

Signed-off-by: Grant Slater <git@firefishy.com>

### DIFF
--- a/cookbooks/rsyncd/recipes/default.rb
+++ b/cookbooks/rsyncd/recipes/default.rb
@@ -49,6 +49,7 @@ systemd_service "rsync-override" do
   service "rsync"
   dropin "override"
   exec_start "/usr/bin/rsync --daemon --no-detach --bwlimit=16384"
+  nice 10
   read_write_paths writable_paths.sort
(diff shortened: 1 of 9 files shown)


cookbooks/rsyncd/recipes/default.rb
```diff
@@ -48,7 +48,8 @@
 systemd_service "rsync-override" do
   service "rsync"
   dropin "override"
-  exec_start "/usr/bin/rsync --daemon --no-detach --bwlimit=16384"
```
tomhughes, line 51: Was getting rid of the bandwidth limit intentional?

Firefishy: yeah it was... Sneaking it in ;-)

Firefishy: Happy to make it a separate PR if desired.

Firefishy: Removed from PR.

PR #535, opened 2022-09-20
Split stateofthemap cookbook to allow parallel testing

Firefishy:
Split the stateofthemap cookbook to allow parallel testing.


tomhughes:
You're missing includes for apache and git as well (which likely used to come via wordpress).

tomhughes:
This won't do any good of course because probably 95% or more of the time is in the wordpress instances which are still altogether and overall this will actually use more resources than before.

PR #536, opened 2022-09-20
Add initial support for other ubuntu ports (eg: arm64)

Firefishy:
This is used for example when running test kitchen on recent Apple devices.

### DIFF
--- a/cookbooks/apt/recipes/default.rb
+++ b/cookbooks/apt/recipes/default.rb
@@ -39,18 +39,26 @@ apt_update "/etc/apt/sources.list" do
   action :nothing
 end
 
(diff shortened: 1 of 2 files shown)


cookbooks/apt/recipes/default.rb
```diff
@@ -39,18 +39,26 @@
   action :nothing
 end
 
-archive_host = if node[:country]
-                 "#{node[:country]}.archive.ubuntu.com"
-               else
-                 "archive.ubuntu.com"
-               end
+if node[:kernel][:machine] == "x86_64"
```
tomhughes, line 42: You could just use `if intel?` here ;-)

PR #537, opened 2022-09-20
Run large test kitchen jobs first in GHA

Firefishy:
Run large jobs first to ensure they to slow down full run completion time.


tomhughes:
:sob:

Firefishy:
Matrix jobs are run in alphabetical order :-( This PR doesn't work.

PR #538, opened 2022-09-21
sotm: use a bundler cache for kitchen runs

Firefishy:
In kitchen cache the `bundle install` from the first jekyll sotm site and re-use the cache for other sotm runs of `bundle install`

### DIFF
--- a/cookbooks/stateofthemap/recipes/jekyll.rb
+++ b/cookbooks/stateofthemap/recipes/jekyll.rb
@@ -39,7 +39,8 @@ apache_module "rewrite"
     depth 1
     user "root"
     group "root"
-    notifies :run, "bundle_install[/srv/#{year}.stateofthemap.org]"
+    notifies :extract, "archive_file[/srv/#{year}.stateofthemap.org/vendor/bundle-tar-cache]", :immediately
+    notifies :run, "bundle_install[/srv/#{year}.stateofthemap.org]", :immediately
   end
 
   directory "/srv/#{year}.stateofthemap.org/_site" do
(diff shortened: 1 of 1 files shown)


Firefishy:
FYI `bundle cache` command only caches the gem sources files and does not cache any compiled `native extension` code. 
jekyll `bundle install` is slow due to https://github.com/sass/sassc-ruby/issues/189

Firefishy:
**20 minutes** fast! :-)

cookbooks/stateofthemap/recipes/jekyll.rb
```diff
@@ -56,15 +57,35 @@
...
     only_if { ::File.exist?("/srv/#{year}.stateofthemap.org/Gemfile") }
   end
 
+  execute "/srv/#{year}.stateofthemap.org/vendor/bundle-tar-cache" do
+    action :nothing
+    command "tar -cf #{Chef::Config[:file_cache_path]}/stateofthemap-gem-bundle-cache.tar bundle"
+    cwd "/srv/#{year}.stateofthemap.org/vendor"
+    only_if { kitchen? }
+    only_if { ::Dir.exist?("/srv/#{year}.stateofthemap.org/vendor/bundle") }
+    not_if { ::File.exist?("#{Chef::Config[:file_cache_path]}/stateofthemap-gem-bundle-cache.tar") }
```
tomhughes, line 86: All these conditions are pointless if the action is `:nothing` because it's not going to run anyway, and when it is run by `notifies` they will be ignored.

cookbooks/stateofthemap/recipes/jekyll.rb
```diff
@@ -56,15 +57,35 @@
...
   end
 
+  archive_file "/srv/#{year}.stateofthemap.org/vendor/bundle-tar-cache" do
+    action :nothing
+    destination "/srv/#{year}.stateofthemap.org/vendor"
+    path "#{Chef::Config[:file_cache_path]}/stateofthemap-gem-bundle-cache.tar"
+    overwrite false
+    only_if { ::File.exist?("#{Chef::Config[:file_cache_path]}/stateofthemap-gem-bundle-cache.tar") }
+    not_if { ::Dir.exist?("/srv/#{year}.stateofthemap.org/vendor/bundle") }
+    only_if { kitchen? }
```
tomhughes, line 67: Why limit it to when running under kitchen? Either it is safe, in which case we can always use it, or it isn't.

Mind if you are caching binaries there would then be questions about invalidating the cache when the dependencies of those binaries change.

tomhughes (changes requested):
Mostly this makes me want to vomit - adding an insane amount of complexity just to speed up the tests a bit.

Firefishy:
Going to abandon this PR. Going to replace SOTM sites with docker containers built upstream.

PR #546, opened 2022-11-11
Update to OpenStreetMap Carto v5.6.2

pnorman:
Fixes a typo causing issues with fonts.


pnorman:
I have chef stopped on the new render server so this is safe to merge. When the new server is caught up, I'll start chef and it will pick up the update.

PR #548, opened 2022-12-05
Test docker buildx

Firefishy:


Firefishy:
Fail. The cache adds no real improvements.

PR #550, opened 2022-12-11
letencrypt: Use default certificate chain

Firefishy:
Use the letsencrypt default certificate chain as recommended by API.

Previously chain of trust needed to be overridden for extended compatibility, I believe this is no longer required.

[Letsencrypt have a complicated chain of trust](https://letsencrypt.org/certificates/), it likely best to use their default recommended chain.


tomhughes:
I will need to investigate this before agreeing as it's a complex area and there were very good reasons for doing what I did at the time.

tomhughes (approved):

PR #551, opened 2022-12-11
NO MERGE - Add tmp inline debug checks for prometheus

Firefishy:
Debug statements for finding reason why prometheus services failing.


tomhughes:
The prometheus issues are fixed now I think, and https://github.com/openstreetmap/chef/commit/8d8fa616f5e5c8933fe5bbe1f1a7dfb548283950 has been added as a permanent version of this to collect logs when a test fails.

PR #552, opened 2022-12-17
replace MultiMaps extension with Kartographer extension

Tigerfell:
I choose Kartographer extension over Maps extension because I could not test the latter in my MediaWiki testing environment. It has previously been problematic as in openstreetmap/operations/issues/249. 

Attribution now looks like 
image

closes openstreetmap/operations#749

Note that the some wiki configuration pages need to be adjusted after the installation in order to make it work properly: `MediaWiki:Kartographer.js`, `MediaWiki:Kartographer-attribution`, `MediaWiki:Kartographer-wv-layer-wikimedia` I can do that.

PR #562, opened 2023-01-17
Attempt to fix missing images in Kartographer wiki extension

Tigerfell:
Following openstreetmap/chef#552, there are still issues with string insertions into the tile URLs. On a Windows computer some zoom factor is inserted into the URL so the extension tries to access `https://tile.openstreetmap.org/12/2053/1365@1.3x.png?lang=en` instead of `https://tile.openstreetmap.org/12/2053/1365.png?lang=en`. AFAIK switching-off high-DPI should fix that.

The second addition configures the extension to stop requesting images for map pins from the tile servers. This behavior was mentioned at the [original proposal](https://github.com/openstreetmap/operations/issues/749) already. I found an previously undocumented setting that will use a static image included in the extension.

### DIFF
--- a/cookbooks/wiki/templates/default/mw-ext-Kartographer.inc.php.erb
+++ b/cookbooks/wiki/templates/default/mw-ext-Kartographer.inc.php.erb
@@ -3,6 +3,7 @@
 
 wfLoadExtension( 'Kartographer' );
 $wgKartographerMapServer = 'https://tile.openstreetmap.org';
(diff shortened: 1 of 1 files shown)


pnorman:
> https://tile.openstreetmap.org/12/2053/1365.png?lang=en

It shouldn't be requesting with lang=en

tomhughes:
I don't think Kartographer has a way to remove that at the moment though it should be harmless as we will ignore it.

PR #563, opened 2023-01-21
minor civicrm upgrade

datendelphin:


tomhughes:
That version doesn't exist:

```
[2023-01-21T11:22:40+00:00] WARN: remote_file[/var/chef/cache/civicrm-5.51.4-wordpress.zip] cannot be downloaded from https://download.civicrm.org/civicrm-5.51.4-wordpress.zip: 404 "Not Found"
```

Will revert this for now.

datendelphin:
Older versions seem paywalled now. I'll check if we could upgrade the major version.
https://civicrm.org/esr

grischard:
Yeah, they charge for the esr releases.

PR #568, opened 2023-02-05
Add Translate extension to Wiki

### LINKED ISSUE #649: Add Translate extension to main wiki
There has been some discussion among the wiki editors whether to add the translate extension. @lectrician1 wrote a [proposal](https://wiki.openstreetmap.org/w/index.php?title=Proposed_features/Add_Translate_extension_to_Wiki&oldid=2344717) that was subsequently accepted. Before the extension can be installed as intended, MediaWiki needs to be updated to version 1.38 (currently 1.37). I can write a PR that adds the extension. 

Includes/obsoletes #555

lectrician1:
This adds the Translate extension to the Wiki. 

All other [extension dependencies](https://www.mediawiki.org/wiki/MediaWiki_Language_Extension_Bundle) have already been installed so no other extensions are added as part of this request.

User rights and the new role of "translateadmin" have been [configured](https://www.mediawiki.org/wiki/Help:Extension:Translate/Configuration) according to the [original proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/Add_Translate_extension_to_Wiki#Configuration).

The OSM ElasticSearch server will now be utilized for storing [translation memories](https://www.mediawiki.org/wiki/Help:Extension:Translate/Translation_memories).

Closes https://github.com/openstreetmap/operations/issues/649

### DIFF
--- a/cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
+++ b/cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
@@ -1,0 +1,36 @@
+<?php
+# DO NOT EDIT - This file is being maintained by Chef
+
+wfLoadExtension( 'Translate' );
+$wgGroupPermissions['user']['translate'] = true;
+$wgGroupPermissions['user']['skipcaptcha'] = true; // Bug 34182: needed with ConfirmEdit
+
+/**
+ * Language code for message documentation. Suggested values are qqq or info.
+ * If set to false (default), message documentation feature is disabled.
+ */
+$wgTranslateDocumentationLanguageCode = 'qqq';
+
+$wgGroupPermissions['autoconfirmed']['translate-messagereview'] = true;
+
+$wgGroupPermissions['translationadmin']['pagetranslation'] = true;
+$wgGroupPermissions['translationadmin']['translate-manage'] = true;
+
+/**
+ * Add a preference "Do not send me email newsletters" in the email preferences.
+ */
+$wgTranslateNewsletterPreference = true;
+
+/** 
+ * Let Translate extension use ElasticSearch to store commonly-used translation messages to suggest to translators
+ */
+$wgTranslateTranslationServices['TTMServer'] = array(
+        'type' => 'ttmserver',
+        'class' => 'ElasticSearchTTMServer',
+        'cutoff' => 0.75,
+        /*
+         * See http://elastica.io/getting-started/installation.html
+         * See https://github.com/ruflin/Elastica/blob/8.x/src/Client.php
+         */
+       'config' => [ 'servers' => [ 'host' => '127.0.0.1', 'port' => 9114 ] ]
+);
--- a/cookbooks/wiki/recipes/default.rb
+++ b/cookbooks/wiki/recipes/default.rb
@@ -122,6 +122,12 @@ mediawiki_extension "Kartographer" do
   template_cookbook "wiki"
 end
 
+mediawiki_extension "Translate" do
+  site "wiki.openstreetmap.org"
+  template "mw-ext-Translate.inc.php.erb"
+  template_cookbook "wiki"
+end
+
 cookbook_file "/srv/wiki.openstreetmap.org/osm_logo_wiki.png" do
   owner node[:mediawiki][:user]
   group node[:mediawiki][:group]


lectrician1:
Could someone please review this? It has been **over a year** since the proposal passed to add this.

Firefishy:
Can you tell me what functionality this introduces?

lectrician1:
@Firefishy Please see [the proposal](https://wiki.openstreetmap.org/wiki/Proposal:Add_Translate_extension_to_Wiki) for an in-depth description about what functionality it offers.

lectrician1:
So...

lectrician1:
I will be. Honestly I have not tested this PR with the current Wiki stack since I don't have chef setup or anything and don't really know how to set it up. I guess creating that docker container for the wiki which has been something I've wanted to do we might think of doing before this as well to test it (openstreetmap/operations#873).

pnorman:
If we install it and then have to remove it in a couple months, what are the consequences? What's the uninstall procedure?

lectrician1:
@pnorman All the pages created as "translated" pages will just be kept in normal page form. For example, Key:Highway and its Spanish translation produced by the Translate extension at Key:Highway/es will become normal pages at Key:Highway and Key:Highway/es. We will need to add the {{Languages}} template to those pages but that should be pretty easy to automate.

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
```diff
@@ -0,0 +1,36 @@
...
+$wgTranslateTranslationServices['TTMServer'] = array(
+        'type' => 'ttmserver',
+        'class' => 'ElasticSearchTTMServer',
+        'cutoff' => 0.75,
+        /*
+         * See http://elastica.io/getting-started/installation.html
+         * See https://github.com/ruflin/Elastica/blob/8.x/src/Client.php
+         */
+       'config' => [ 'servers' => [ 'host' => '127.0.0.1', 'port' => 9114 ] ]
+);
```
lens0021, line 36:
```suggestion
$wgTranslateTranslationServices['TTMServer'] = [
        'type' => 'ttmserver',
        'class' => 'ElasticSearchTTMServer',
        'cutoff' => 0.75,
        /*
         * See http://elastica.io/getting-started/installation.html
         * See https://github.com/ruflin/Elastica/blob/8.x/src/Client.php
         */
       'config' => [ 'servers' => [ 'host' => '127.0.0.1', 'port' => 9114 ] ]
];
```
The mix use of old syntax and new syntax has slightly confused me :)

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
```diff
@@ -0,0 +1,36 @@
+<?php
+# DO NOT EDIT - This file is being maintained by Chef
+
+wfLoadExtension( 'Translate' );
+$wgGroupPermissions['user']['translate'] = true;
+$wgGroupPermissions['user']['skipcaptcha'] = true; // Bug 34182: needed with ConfirmEdit
```
lens0021, line 6: Bugzilla 34182 is https://phabricator.wikimedia.org/T36182

lens0021: Though the bug, I am not sure if it is good to disable the captcha for all logged-in users.
In another perspective, does OSM wiki have ConfirmEdit enabled? (Is this line needed?)
Anyway, I think it would be good if 'autoconfirmed' is used than 'user'.

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
```diff
@@ -0,0 +1,36 @@
...
+$wgGroupPermissions['autoconfirmed']['translate-messagereview'] = true;
+
+$wgGroupPermissions['translationadmin']['pagetranslation'] = true;
+$wgGroupPermissions['translationadmin']['translate-manage'] = true;
+
+/**
+ * Add a preference "Do not send me email newsletters" in the email preferences.
+ */
+$wgTranslateNewsletterPreference = true;
+
```
lens0021, line 23:
```suggestion
/**
 * Add a preference "Do not send me email newsletters" in the email preferences.
 */
$wgTranslateNewsletterPreference = true;

```
This configuration variable removed in 2020. (MediaWiki 1.35)

https://github.com/wikimedia/mediawiki-extensions-Translate/commit/1c43244d7c56a24fe95cde1dec9a94cdfeceae69

lens0021:
> Honestly I have not tested this PR with the current Wiki stack since I don't have chef setup or anything and don't really know how to set it up.

@lectrician1 If you can, could you please take the time to test it with Chef?

> If we install it and then have to remove it in a couple months, what are the consequences? What's the uninstall procedure?

@Nikerabbit I am sorry to mention you suddenly. Do you know what will happen if Translate extension is uninstalled? Are there any cases of this happened on any wiki?

Nikerabbit:
Not to my knowledge. Main changes would be <translate> tags would displaying verbatim and all translations would become editable as normal pages.

lectrician1:
@lens0021 I do not have a local dev environment for the chef configuration. Is it that hard to setup? The configuration looks very osm-server specific.

Could someone else possibly test the configuration?

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
tomhughes: How exactly does a translation extension involve sending newsletters? If adding this extension is going to result is us sending bulk emails then we're going to need a clear understanding of what it will be doing and the legal/privacy implications...

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
tomhughes: Yes we have `ConfirmEdit` installed using hcaptcha.

cookbooks/wiki/templates/default/mw-ext-Translate.inc.php.erb
lens0021: You don't have to worry about it because the code does not exist anymore. It seems the configuration only provided a checkbox on the preference page. So maybe the publisher of the newsletter had manually sent mails? I don't know.

1ec5:
Yes, I do think we want the Translate extension. For now the translation administrators would only enable it for stuff other than tag description pages.

@tomhughes seems to have adapted this PR for the test wiki in 0cd4d0d88b13cd09359402297508bc16a1c7e58d and bbb97ede0cf9a1453381ed4fb7972e3b7742078f. Unfortunately we hit an error in Special:SearchTranslations that hampered one of the key features, translation memory. It might have to do with WikibaseCirrusSearch (openstreetmap/operations#903), which is also being tested on the test wiki simultaneously. I don’t think Translate actually depends on WikibaseCirrusSearch; we just ended up testing both extensions in tandem.

PR #570, opened 2023-02-11
Add docker based welcome.openstreetmap.org using podman

tomhughes:
This is an alternative to #566 that adds support for running container services using podman along with modified versions of @Firefishy's work to run welcome.openstreetmap.org as a container.

Unfortunately https://github.com/containers/podman/issues/12778 means it's not currently possible to use `DynamicUser` to run each container as a separate user so instead it edits `/etc/subuid` and `/etc/subgid` to reserve a block of user IDs for containers and then uses `--userns=auto` to allocate each container a group of IDs from that block.

It also switches networking to use `slirp4netns` mode (which would be the default if we were running as a non-root user) which is probably a bit less efficient but avoids needing to add firewall rules which then cause issues with shorewall.


Firefishy:
I looked into `slirp4netns` earlier. The performance is quite good, especially if you match the MTU to the upstream interface (1500). We can always optimise later.

Firefishy (approved):

PR #571, opened 2023-02-11
sotm: Switch to using containers for jekyll sites

Firefishy:
Switch from chef managed jekyll install and build, to using upstream container versions of the sites.


cookbooks/stateofthemap/templates/default/apache.jekyll.erb
```diff
@@ -14,7 +14,9 @@
 
 <VirtualHost *:443>
         ServerName <%= @year %>.stateofthemap.com
-        ServerAlias <%= @year %>.sotm.org
+        <% @aliases.each do |alias_name| -%>
+          ServerAlias <%= alias_name %>
+       <% end -%>
```
tomhughes, line 19: What's the reason for moving the aliases out into the recipe? It seems inconsistent to construct the primary name here but not the aliases?

Also the indentation here is inconsistent.

cookbooks/stateofthemap/templates/default/apache.jekyll.erb
```diff
@@ -38,14 +40,9 @@
         SSLCertificateFile /etc/ssl/certs/<%= @year %>.stateofthemap.org.pem
         SSLCertificateKeyFile /etc/ssl/private/<%= @year %>.stateofthemap.org.key
 
-        DocumentRoot /srv/<%= @year %>.stateofthemap.org/_site
+        RequestHeader set X-Forwarded-Proto “https”
+        RequestHeader set X-Forwarded-Port “443”
```
tomhughes, line 44: Why are these needed?

tomhughes (changes requested):

cookbooks/stateofthemap/templates/default/apache.jekyll.erb
Firefishy: Updated to be consistent with alias usage.

cookbooks/stateofthemap/templates/default/apache.jekyll.erb
Firefishy: While this isn't used by the sotm or welcome containers, it is used to let the container backend know that the request is an HTTPS request. It might be desired in future for a backend to redirect / reject the request if it is not an HTTPS request.

tomhughes:
Merged.

PR #572, opened 2023-02-12
Remove tile per IP rate limiting

pnorman:
With the ability to ratelimit on the CDN, this is no longer necessary.

This change does not clean up the existing `/srv/tile.openstreetmap.org/conf` directory and other files. How can I do that within chef?


tomhughes:
Just change the directory block to:

```
directory "/srv/tile.openstreetmap.org/conf" do
  action :delete
  recursive true
end
```

and then once it has had a chance to run everywhere you can remove that.

cookbooks/tile/recipes/default.rb
```diff
@@ -546,31 +528,6 @@
   files_mode "755"
 end
 
-template "/usr/local/bin/tile-ratelimit" do
```
Firefishy, line 549: This should likely be replaced by:
```
file "/usr/local/bin/tile-ratelimit" do
  action :delete
end
```

cookbooks/tile/recipes/default.rb
```diff
@@ -546,31 +528,6 @@
...
-  user "tile"
-  group "adm"
-  exec_start "/usr/local/bin/tile-ratelimit"
-  nice 10
-  sandbox true
-  read_write_paths "/srv/tile.openstreetmap.org/conf"
-  restart "on-failure"
-end
-
-service "tile-ratelimit" do
```
Firefishy, line 568:
```
service "tile-ratelimit" do
  action [:stop, : disable]
end
```

cookbooks/tile/recipes/default.rb
```diff
@@ -546,31 +528,6 @@
...
 end
 
-template "/usr/local/bin/tile-ratelimit" do
-  source "tile-ratelimit.erb"
-  owner "root"
-  group "root"
-  mode "755"
-end
-
-systemd_service "tile-ratelimit" do
```
Firefishy, line 556: Change the order to stop and disable the service, then delete it.
```
systemd_service "tile-ratelimit" do
  action :delete
end
```

PR #574, opened 2023-02-12
Clarify licensing and improve styles on planet

russss:
Apparently there is [some room for confusion](https://lists.openstreetmap.org/pipermail/talk/2023-February/088042.html) about how the data is licensed... This changeset tries to clarify the wording on the planet front page a bit, as I think it could be clearer.

I also took the opportunity to modernise the CSS and make it less of a disaster on mobile devices (not that planet is much use on a mobile device, but at least this should now be obvious quicker) and do a bit of copy-editing.

Obviously difficult to test, but here's the page locally with the dynamic stuff subbed in - broken images in the directory listing are due to this:

## Mobile
(It's not visible here but you can horizontal-scroll the directory listing)


tomhughes:
Please don't mix up content changes and style changes in the one commit - it makes it almost impossible to tell what changes have been made to the content and we probably need to get LWG to sign off on those.

russss:
Fair enough - I'll put this PR on hold and do another for the content changes first.

russss:
Superseded by #646

PR #576, opened 2023-02-12
Clarify licensing wording on planet.osm.org

russss:
This contains only the content changes from #574 - a bit more info and screenshots of the updated wording can be found in that pull request.

The intention of this change is to make more explicit what license the data is distributed under, as this has apparently [caused confusion](https://lists.openstreetmap.org/pipermail/talk/2023-February/088042.html). It also adds the "Copyright OpenStreetMap" attribution link which I believe is technically needed here too.

It also includes some minor copy-editing.

(This change alone causes a slight layout glitch as the OSM logo overlaps the warning box. This is fixed in #574.)

### DIFF
--- a/cookbooks/planet/files/default/cgi/HEADER.cgi
+++ b/cookbooks/planet/files/default/cgi/HEADER.cgi
@@ -67,8 +67,7 @@ print("""
 <h1>Planet OSM</h1>
 
 <p>
-The files found here are regularly-updated, complete copies of the OpenStreetMap.org
-database, and those published before the 12 September 2012 are distributed under a Creative Commons Attribution-ShareAlike 2.0 license, those published after are  Open Data Commons Open Database License 1.0 licensed. For more information, <a href="https://wiki.openstreetmap.org/wiki/Planet.osm">see the project wiki</a>.
+The files found here are regularly-updated, complete copies of the <a href="https://openstreetmap.org">OpenStreetMap</a> database.
 </p>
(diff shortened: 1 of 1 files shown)


cookbooks/planet/files/default/cgi/HEADER.cgi
```diff
@@ -90,32 +89,42 @@ database, and those published before the 12 September 2012 are distributed under
...
         A smaller file with complete metadata for all changes ('changesets') in
         XML format is also available.
         </p>
     </td>
     <td>
         <p>
-        You are granted permission to use OpenStreetMap data by
-        <a href="https://osm.org/copyright">the OpenStreetMap License</a>, which also describes
-        your obligations.
+        This open data is <a href="https://osm.org/copyright">copyright &copy; OpenStreetMap</a>.
```
tomhughes, line 105: I think this is the only substantive difference, as against rearranging or rewording of existing text.

I also think it's problematic because it's asserting that "OpenStreetMap" owns the copyright but there is no such entity and even the foundation is not the copyright holder (to the extent that copyright even matters as things like database rights are likely more important) as user's do not assign copyright in their contributions.

tomhughes (changes requested):

cookbooks/planet/files/default/cgi/HEADER.cgi
russss: The [recommended attribution text](https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines#Attribution_text) now is just "OpenStreetMap", although it seems the "©" doesn't seem to be recommended now.

Maybe just "This open data is provided by OpenStreetMap", linking back to osm.org/copyright?

tomhughes: Attribution is not the same as claiming to own the copyright of course.

russss: Yeah, unless I'm mistaken, the "copyright OpenStreetMap" wording was recommended for a while but it looks like that's changed.

I've tweaked the wording now.

russss:
I may re-submit this against the updated HTML if #646 is merged.

PR #578, opened 2023-02-12
switch2osm: Use container

Firefishy:
Signed-off-by: Grant Slater <github@firefishy.com>


tomhughes (changes requested):
You need to move the switch2osm role from ridley to naga...

PR #580, opened 2023-02-14
Introduce podman_site resource

tomhughes:
This introduces a `podman_site` resource to handle the boilerplate of allocating host port numbers and arranging an apache proxy to the podman container.


Firefishy:
Looks good to me.

PR #588, opened 2023-05-06
Atomically update taginfo databases

joto:
Make sure we always see one version of each taginfo database.


tomhughes:
I don't it actually needs to be that complex - just changing the first `mv` to `ln` should have the same effect?

joto:
You are probably right. We will need a `-f` on the `ln` though if we don't remove the files in `old` beforehand.

I was thinking that our problem might have come from `mv` doing something strange like removing all targets first and then only `rename`ing the files, that's why I originally came up with that loop. But I just checked and `mv` isn't doing that, so your solution is better.

PR #590, opened 2023-06-15
Enable IPv6 in dokken tests

Firefishy:


Firefishy:
Annoyingly it seems npm (and node) use IPv6 addresses by default over IPv4 addresses and does not fallback.

* `node --network-family-autoselection` is available in very recent node versions to enable [Happy Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs), but requires [bleeding edge fixes](https://github.com/nodejs/node/pull/48255) to work properly.
* Alternative is to change precedence back to IPv4 eg: `node --dns-result-order=ipv4first "$(which npm)"`

tomhughes:
I think the solution is to use a ULA adddress instead of `2001:db8` as ULA addresses are treated as lower priority than IPv4 when resolving names. I've added a commit to test that out...

PR #591, opened 2023-06-15
wordpress: add 2fa and last-login plugins

Firefishy:
* Add ability to activate plugins via cli
* Add 2fa and last-login plugins


Firefishy:
I will fix this PR by added the missing `wp core install --url=example.com --title=Example --admin_user=supervisor --admin_password=strongpassword --admin_email=info@example.com
` logic required for test kitchen.

PR #592, opened 2023-06-20
networking: allow opportunistic DNS Over TLS

Firefishy:
DNS over TLS seems to be well supported in systemd-resolved

In limited testing DNS over TLS seems to work more consistently than standard DNS when IPv6 and/or DNSSEC are not functioning correctly (DNSSEC fallback).


tomhughes:
What do  you mean by "when not functioning correctly" exactly? What sort of failures are you seeing? Why should the transport change that?

What about machines that aren't using google/cloudflare as servers? Will this have a negative impact when the server doesn't have DoT support?

Firefishy:
> What do you mean by "when not functioning correctly" exactly? What sort of failures are you seeing? Why should the transport change that?
> 

When IPv6 is enabled within the container, but connections over IPv6 timeout (eg: GitHub Actions or Docker on Mac) because IPv6 is not routable on the host system.

> What about machines that aren't using google/cloudflare as servers? Will this have a negative impact when the server doesn't have DoT support?

It will not enable DNSoverTLS for servers not definite in `a.b.c.d#example.com eg: 1.1.1.1#cloudflare-dns.com` format.
It should therefore have no affect on our systems which use other DNS.

tomhughes:
Oh I thought the hashes were just comments, not part of the syntax ;-)

I'm still working on trying to resolve the IPv6 in containers thing properly though...

Firefishy:
@tomhughes OK with me merging this? I have disabled in <22.04 as the support doesn't seem as mature in that version of systemd-resolved

PR #593, opened 2023-06-23
Specify DNS server to use for dokken

Firefishy:
Attempt to fix the broken nested docker test for community

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -8,6 +8,11 @@ driver:
     - container=dokken
   ipv6: true
   ipv6_subnet: "fd95:8552:964c::/64"
+  dns:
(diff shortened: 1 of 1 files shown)


Firefishy:
This seems to fix the docker in docker (dokken) community test.

Firefishy:
Likely an alternative would be to disable IPv6 within the docker in docker by removing these lines: https://github.com/openstreetmap/chef/blob/c62e005dbcf218ec19734b8b436ae8ddf82d3691/cookbooks/docker/templates/default/daemon.json.erb#L8-L11

tomhughes:
Yes but we need that to get logging of IPv6 addresses in discourse.

PR #594, opened 2023-07-03
Add awscli cookbook

Firefishy:
Add `awscli` cookbook.

The distro versions of aws cli not supported by AWS. Some features only work reliably with newer releases eg: `aws s3 sync`

The cookbook will be used later.

### DIFF
--- a/cookbooks/awscli/metadata.rb
+++ b/cookbooks/awscli/metadata.rb
@@ -1,0 +1,9 @@
+name              "awscli"
+maintainer        "OpenStreetMap Administrators"
+maintainer_email  "admins@openstreetmap.org"
+license           "Apache-2.0"
+description       "Installs and configures awscli"
+
+version           "1.0.0"
+supports          "debian"
+supports          "ubuntu"
--- a/cookbooks/awscli/recipes/default.rb
+++ b/cookbooks/awscli/recipes/default.rb
@@ -1,0 +1,89 @@
+#
+# Cookbook:: awscli
+# Recipe:: default
+#
+# Copyright:: 2023, OpenStreetMap Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


Firefishy:
Confirmed working with ubuntu-20.04, ubuntu 22.04 and debian-12.

cookbooks/awscli/metadata.rb
```diff
@@ -0,0 +1,8 @@
+name              "awscli"
+maintainer        "OpenStreetMap Administrators"
+maintainer_email  "admins@openstreetmap.org"
+license           "Apache-2.0"
+description       "Installs and configures awscli"
+
+version           "1.0.0"
+supports          "ubuntu"
```
tomhughes, line 8: Not sure how it work on debian with that? I thought chef would reject it...

cookbooks/awscli/recipes/default.rb
```diff
@@ -0,0 +1,74 @@
...
+archive_file "#{cache_dir}/#{awscli_zip}" do
+  path "#{cache_dir}/#{awscli_zip}"
+  destination "#{cache_dir}/awscli"
+  strip_components 1
+  action :nothing
+  subscribes :extract, "remote_file[#{cache_dir}/#{awscli_zip}]", :immediately
+end
+
+execute "install-#{awscli_zip}" do
+  command "./install --bin-dir /opt/awscli/bin --install-dir /opt/awscli --update"
```
tomhughes, line 70: I really don't like running upstream install scripts if there is any other option - you have no idea what they are going to to.

Does it actually do anything that couldn't be achieved just by copying the files?

cookbooks/awscli/metadata.rb
Firefishy: The kitchen tests did pass. Weird. Fixed. now.

cookbooks/awscli/recipes/default.rb
Firefishy: yeah, I am not super excited about the install script. It isn't super complicated what it does: extracts version by running `aws --version` from package, `cp -r` into install-dir directory and updates some symlinks.

I guess I could replicate that in chef.

Firefishy: I have now replaced the installer with a custom `ruby_block` which mimics the behaviour of the installer.

I initially tried using a `ruby_block` to only extract the version `aws --version` and then lazy inject those variables into chef resources, but that was overly comples.

PR #597, opened 2023-08-07
Pull image during podman service startup

Firefishy:
`ghcr.io` has occasionally been returning "502 - Bad Gateway" errors which causes the Github Action kitchen tests to fail.

I have been able to occasionally replicate these failures locally.

Adding in a `podman pull` in the service `exec_start_pre` resolved the issue for me. I believe `pull` has automatic retry logic.

~~Also added a few basic podman tests~~ Will add these in another PR.

### DIFF
--- a/cookbooks/podman/resources/service.rb
+++ b/cookbooks/podman/resources/service.rb
@@ -32,7 +32,11 @@ action :create do
     type "notify"
     notify_access "all"
     environment "PODMAN_SYSTEMD_UNIT" => "%n"
-    exec_start_pre "/bin/rm --force %t/%n.ctr-id"
+    # The exec_start_pre "podman pull" is to workaround the occasional 502 error returned from the registry.
+    exec_start_pre [
+      "/bin/rm --force %t/%n.ctr-id",
+      "-/usr/bin/podman pull #{new_resource.image}"
+    ]
     exec_start "/usr/bin/podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --userns=auto --label=io.containers.autoupdate=registry --network=slirp4netns:mtu=1500 #{publish_options} --rm --sdnotify=conmon --detach --replace --name=%N #{new_resource.image}"
     exec_stop "/usr/bin/podman stop --ignore --time=10 --cidfile=%t/%n.ctr-id"
     exec_stop_post "/usr/bin/podman rm --force --ignore --cidfile=%t/%n.ctr-id"


cookbooks/podman/resources/service.rb
```diff
@@ -32,7 +32,11 @@
     type "notify"
     notify_access "all"
     environment "PODMAN_SYSTEMD_UNIT" => "%n"
-    exec_start_pre "/bin/rm --force %t/%n.ctr-id"
+    # The exec_start_pre "podman pull" is to workaround the occasional 502 error returned from the registry.
+    exec_start_pre [
+        "/bin/rm --force %t/%n.ctr-id",
+        "-/usr/bin/podman pull #{new_resource.image}"
```
tomhughes, line 38: Why would the pull not just trigger the same 502 errors? Of course it probably is but because you're suppressed errors you wont know when it does!

cookbooks/podman/resources/service.rb
```diff
@@ -32,7 +32,11 @@
...
+      ]
```
tomhughes, line 39: I'm amazed cookstyle isn't complaining about this indentation.

Firefishy: Fixed the formatting.

cookbooks/podman/resources/service.rb
Firefishy: The 502 errors seem sporadic from GHCR.io.

The new code effectively just adds a retry by pulling the image first, using the fail safe `-/usr/bin/podman pull...` if this pull fails it is ignored, which is then reliant on the `podman run` triggering the pull.

Before this code I was able to trigger a failure at least 1 in 10. With the addition I was unable to get it to fail.

podman supposedly has built in `podman pull` image layer retry logic, but I wasn't able to figure it out.

Further investigation could maybe reveal if `ghcr.io` limit simultaneous connection

tomhughes: Image pulling does retry three times by default (https://github.com/containers/common/blob/main/libimage/copier.go#L27) but at one second intervals which probably isn't very helpful.

I assume that will apply to the implicit pull in a run as much as an explicit one so I assume all we're achieving here is doubling the number of retries by having three in the pull and then three more in the run - plus we may get a slightly longer delay between the two groups.

tomhughes: Actually it looks like it only retries when [IfErrrorRetryable](https://github.com/containers/common/blob/b13df748a2b8929e7f4c4edaf6a2bd159157c0cb/pkg/retry/retry.go#L63) says it's retryable and I think that only includes network errors and not HTTP level errors so likely this just effectively adds a single retry.

Firefishy: The errors seem to occasionally be DNS too:

```
Aug 08 03:46:44 dokken systemd[1]: Starting stateofthemap.org.service - Container service for stateofthemap.org...
Aug 08 03:46:44 dokken podman[3784]: 2023-08-08 03:46:44.400044637 +0000 UTC m=+0.079591918 system refresh
Aug 08 03:46:44 dokken podman[3784]: Trying to pull ghcr.io/openstreetmap/stateofthemap-website:latest...
Aug 08 03:46:49 dokken podman[3784]: Error: initializing source docker://ghcr.io/openstreetmap/stateofthemap-website:latest: pinging container registry ghcr.io: Get "https://ghcr.io/v2/": dial tcp: lookup ghcr.io: no such host
Aug 08 03:46:49 dokken systemd[1]: stateofthemap.org.service: Main process exited, code=exited, status=125/n/a
Aug 08 03:46:49 dokken podman[3819]: Error: reading CIDFile: open /run/stateofthemap.org.service.ctr-id: no such file or directory
Aug 08 03:46:49 dokken systemd[1]: stateofthemap.org.service: Control process exited, code=exited, status=125/n/a
Aug 08 03:46:49 dokken systemd[1]: stateofthemap.org.service: Failed with result 'exit-code'.
Aug 08 03:46:49 dokken systemd[1]: Failed to start stateofthemap.org.service - Container service for stateofthemap.org.
Aug 08 03:46:50 dokken systemd[1]: stateofthemap.org.service: Scheduled restart job, restart counter is at 1.
Aug 08 03:46:50 dokken systemd[1]: Stopped stateofthemap.org.service - Container service for stateofthemap.org.
Aug 08 03:46:50 dokken systemd[1]: Starting stateofthemap.org.service - Container service for stateofthemap.org...
Aug 08 03:46:50 dokken podman[3836]: Trying to pull ghcr.io/openstreetmap/stateofthemap-website:latest...
Aug 08 03:46:51 dokken podman[3836]: Getting image source signatures
Aug 08 03:46:51 dokken podman[3836]: Copying blob sha256:1d519f69af07c9cb256b737adb0e73c16e357862a7bef9b7603dcbc5178c2a36
```

PR #599, opened 2023-08-09
Switch to using inspec for testing

tomhughes:


Firefishy:
Wow. Looks great to me.

PR #600, opened 2023-08-11
networking: allow opportunistic DNS Over TLS

Firefishy:
Bringing back this PR. Without is most kitchen runs fail locally for me due to DNS failure.

DNS over TLS seems to be well supported in recent systemd-resolved

In limited testing DNS over TLS seems to work more consistently than standard DNS when IPv6 and/or DNSSEC are not functioning correctly (DNSSEC fallback).

### DIFF
--- a/cookbooks/networking/attributes/default.rb
+++ b/cookbooks/networking/attributes/default.rb
@@ -8,9 +8,18 @@ default[:networking][:firewall][:http_connection_limit] = nil
 default[:networking][:firewall][:allowlist] = []
 default[:networking][:roles] = {}
 default[:networking][:interfaces] = {}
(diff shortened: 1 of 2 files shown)


tomhughes (commented):
I'm not really sure how I'm supposed to review something when nobody can explain to me why it helps.

Of course from a practical point of view this will do nothing on most of our production systems as they won't be using the default resolver list.

Mostly I'm just fed up with people making constant random cargo cult changes to resolver configuration which just wind up breaking as much as they fix.

Firefishy:
Correct for calling me out on this, it is cargo culting in fix an issue without understanding what is causing the problem.

PR #606, opened 2023-08-24
Add test documentation

Firefishy:
Adds documentation for installing the test kitchen requirements and then running the tests locally.

I've recommended the use of rbenv as it adds maximum compatibility across Linux and MacOS platforms. Added a note that rbenv isn't strictly required.


CONTRIBUTING.md
```diff
@@ -9,3 +9,57 @@ https://help.github.com/articles/using-pull-requests
...
+
+## Running the Infrastructure Tests locally
+
+- **[Cookstyle](https://docs.chef.io/workstation/cookstyle/)** is used for linting, ensuring that our Chef recipes follow style guidelines and best practices.
+- **[Test Kitchen](https://kitchen.ci/)** combined with **InSpec** and [Dokken](https://github.com/test-kitchen/kitchen-dokken) is used to verify the functionality of our Chef code, ensuring it behaves as expected.
+
+The following guidelines are to help set up and run these checks locally:
+
+#### **1. Install Docker**
+- Visit [Docker's official site](https://www.docker.com/products/docker-desktop) to download and install Docker Desktop for Mac.
```
tomhughes, line 21: Maybe remove `for Mac` here if this is meant to cover linux as well? or do we need separate instructions for linux here?

CONTRIBUTING.md
```diff
@@ -9,3 +9,57 @@ https://help.github.com/articles/using-pull-requests
...
+#### **4. Install Required Ruby Version (recommended)**
+Navigate to the git checkout of the OpenStreetMap chef repo and run:
+```bash
+rbenv install
+```
+This will install the recommended version of ruby for running the tests. The recommended version of ruby is defined in the [.ruby-version](.ruby-version) file.
+
+#### **5. Install Dependencies with Bundler**
+```bash
+rbenv exec gem install bundler
```
gravitystorm, line 42: There's no need to `rbenv exec` everything here. `rbenv` puts `.rbenv/shims` onto the front of your path, which includes shims for `bundler`, `gem` etc. So `gem install bundler` and `bundle exec cookstyle` will work fine without `rbenv exec`.

Firefishy: @gravitystorm Happy with the explanation? If you feel strongly, then I don't mind modifying the documentation.

PR #610, opened 2023-08-28
Update yarnpkg APT signing key

pnorman:
`apt update` was complaining about an expired key. I checked the setup instructions and found the key has changed, and got the new one with `curl -s https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --show-keys --with-colons - | awk -F':' '$1=="pub"{print $5}'`

I validated this from two machines for paranoia.


Firefishy (approved):

Firefishy:
`curl -sL https://dl.yarnpkg.com/debian/pubkey.gpg | gpg --keyid-format long` confirms correct key.

Another option would be to update to use the URL for the key, instead of hardcoding.

PR #611, opened 2023-08-31
Use GHCR dokken images in test kitchen

Firefishy:
Switch to GHCR.io dokken images for improved image download latency and to avoid potential docker hub access limits.

### DIFF
--- a/.github/workflows/test-kitchen.yml
+++ b/.github/workflows/test-kitchen.yml
@@ -128,6 +128,12 @@ jobs:
             os: ubuntu-2204
       fail-fast: false
     steps:
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v2
+      with:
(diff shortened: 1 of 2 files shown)


tomhughes:
Why do we need to login if we're only reading? I thought that was only necessary to push or does it give us higher rate limits or something?

Firefishy:
> Why do we need to login if we're only reading? I thought that was only necessary to push or does it give us higher rate limits or something?

Without the login to GHCR all pull requests within the GHCR primary runner fail. [See example](https://github.com/openstreetmap/chef/actions/runs/5989754274). Logging-In with readonly permissions seems to fix the error. The [docs do say if downloading via your GITHUB_TOKEN the downloads are unlimited](https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages).

PR #613, opened 2023-09-11
Set default effective_io_concurrency and random_page_cost to SSD values

pnorman:
Since all drives are SSDs in arrays with at least two drives, 256 is a reasonable default. NVMe drives could have this set even higher, but this is a reasonable start.


tomhughes:
So far we've stuck with keeping the defaults equal to the upstream defaults and only overriding them as needed but I guess we don't have to.

pnorman:
We've overridden this on all the nominatim machines and api DB servers, and I was about to do so for all the render servers, so I figure we might as well do it globally. Some defaults just aren't sensible on a server of reasonable capacity, since PostgreSQL aims for defaults that will work on a very small machine.

tomhughes:
Shouldn't this also remove the overrides that we already have in place then?

pnorman:
I've removed the overrides that were 256 or near it. Nominatim has per-server specific overrides, so I left those in place. cc @lonvia 

I also added random_page_cost since we were overwriting it on tile servers, all the api DB servers, and nominatim servers. I left the nominatim override in place, since that's been tuned to unusual values.

PR #614, opened 2023-09-12
Set reasonable PostgreSQL GUCs for tile servers

pnorman:
These are based off of the osm2pgsql.org recommendations, scaled for the larger servers we run, as well as experience.

This should be merged **after** #613 and apply cleanly. If it doesn't apply cleanly, I'll rebase.


tomhughes:
What is a GUC exactly?

roles/tile.rb
```diff
@@ -37,16 +37,17 @@
     :settings => {
       :defaults => {
         :max_connections => "250",
-        :temp_buffers => "32MB",
+        :shared_buffers => "16GB",
```
tomhughes, line 40: This is currently configured on a per-server basis, though I suspect some may have been copy and pasted without paying attention to the amount of memory in a new server.

roles/tile.rb
```diff
@@ -37,16 +37,17 @@
...
         :work_mem => "128MB",
+        :maintenance_work_mem => "8GB",
```
tomhughes, line 42: This is also currently configured on a per-server basis.

roles/tile.rb
pnorman: I'd like to have a sensible default. Do you think I should remove the other settings of it where it's close to this?

roles/tile.rb
pnorman: We've got cases where it's configured both on a role level and a server level. I'll remove the setting of it for many of the servers where the defaults make sense.

pnorman:
> What is a GUC exactly?

PostgreSQL [configuration parameter](https://www.postgresql.org/docs/current/config-setting.html)

roles/tile.rb
tomhughes: I think it was originally done as a proportion of memory although I as say I'm not sure it hasn't sometimes been copy and pasted instead of recalculated for new servers.

Probably machines these days always have sufficiently vast amounts of memory that we can just use a single value.

roles/balerion.rb
```diff
@@ -21,8 +21,6 @@
   :postgresql => {
     :settings => {
       :defaults => {
-        :shared_buffers => "8GB",
-        :maintenance_work_mem => "7144MB",
```
tomhughes, line 25: If we're going to drop this from the per-server configuration do we not want to add a shared setting? Otherwise it will drop to the 64Mb default?

pnorman: Looks like a rebase mistake wiped out that change

pnorman: Looks like a rebase mistake wiped out that change

PR #618, opened 2023-09-18
Add allow wiki main page on root + osmf site url change

Firefishy:
This change changes https://wiki.osmfoundation.org/ site to https://osmfoundation.org/

Allows / to serve /wiki/Main_Page without a redirect.


Firefishy:
There will be a brief self-signed certificate used when this goes live as we'd need to issue a new certificate.

PR #627, opened 2023-09-28
render: Run lowzoom daily

Firefishy:
Now that renderd is significantly faster, it becomes practical to run lowzoom updates daily.

### DIFF
--- a/cookbooks/tile/recipes/default.rb
+++ b/cookbooks/tile/recipes/default.rb
@@ -614,7 +614,7 @@ end
 
 systemd_timer "render-lowzoom" do
   description "Render low zoom tiles"
-  on_calendar "Fri *-*-* 23:00:00 UTC"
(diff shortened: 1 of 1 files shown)


pnorman:
I'm a bit skeptical on this given our slowest machines take 4.5h to do the re-render. Would it be better to adjust the low/mid zoom breakpoint?

Firefishy:
@pnorman @tomhughes Any protest if I merge this as a trial? We can revert to adapt as desired based on outcome.

tomhughes:
I'd be concerned about merging while we still have issues with OOM during low zoom renders.

pnorman:
With pyrene gone our slowest machines are faster, but still take 4 hours to do a re-render. I don't think we should do this every day.

Firefishy:
I have merged it for now. We can revert if it causes any issues.

PR #629, opened 2023-10-06
wordpress: unpin wp-fail2ban release

Firefishy:


grischard:
From IRC:

> The issue with the wp-fail2ban was that the latest release was broken. I pinned to 5.1.1 as per commit message. I should have added more detail. Can now likely unpin wp-fail2ban. I'll PR it.

Do you remember how it was broken? Do we have a ticket?

PR #632, opened 2023-10-21
Use dual stack (IPv4/IPv6) S3 endpoints for public-facing planet

grischard:
Fixes https://github.com/openstreetmap/operations/issues/983


tomhughes:
Does this also fix the other issue that @Firefishy was talking about with the bucket name?

Firefishy:
> Does this also fix the other issue that @Firefishy was talking about with the bucket name?

I'm unsure what the issue was.

Firefishy:
@grischard I am not super eager for the double lookup, but unfortunately the dualstack URL doesn't support the short form I used.

If the 2 look ups somehow returned different results (not that I can see how), the result from AWS S3 is a `IllegalLocationConstraintException` error. `<Message>The xxx location constraint is incompatible for the region specific endpoint this request was sent to.`

tomhughes:
Oh sorry it was @pnorman that raised it not you...

grischard:
@Firefishy removed double call in latest commit

Firefishy:
> @Firefishy removed double call in latest commit

Yes that avoids the double call by making every request, even those not redirecting, wait for a response from the lookup script. The lookups are serial and protected from concurrency by locks.

Realistically likely minimal impact.

Firefishy:
Replaced by https://github.com/openstreetmap/chef/pull/633

PR #634, opened 2023-10-27
podman: avoid starting and immediate restart on first run

Firefishy:
PR to avoid starting the service and then restarting it on first run.

Separate out the start and use service status to make it a "noop" if service is already running due to restart.

### DIFF
--- a/cookbooks/podman/resources/service.rb
+++ b/cookbooks/podman/resources/service.rb
@@ -41,9 +41,16 @@ action :create do
     restart "on-failure"
   end
 
+  # No action :start here to avoid a start and then immediate :restart, due to subscribe, on first run
   service new_resource.service do
-    action [:enable, :start]
-    subscribes :restart, "systemd_service[#{new_resource.service}]"
+    action :enable
+    subscribes :restart, "systemd_service[#{new_resource.service}]", :immediately
+  end
+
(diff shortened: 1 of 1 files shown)


tomhughes (commented):
Is just adding `:immediately` to the subscribe not enough? That way it will happen immediately while the start will be delayed to the end of the run by time the restart will already have happened? In fact you could also change it to `:try_restart` so that it does nothing if the service isn't running yet.

Firefishy:
With this...
```
  service new_resource.service do
    action [:enable, :start]
    supports :status => true, :restart => true
    subscribes :restart, "systemd_service[#{new_resource.service}]", :immediately
  end
```
On a first run the actions run in order `:enable`, `:start` (1st start) and then `:restart` (2nd start).

cookbooks/podman/resources/service.rb
```diff
@@ -41,9 +41,17 @@
     restart "on-failure"
   end
 
+  # No start action here to avoid a start and then immediate restart on first run
   service new_resource.service do
-    action [:enable, :start]
-    subscribes :restart, "systemd_service[#{new_resource.service}]"
+    action :enable
+    supports :status => true, :restart => true
```
tomhughes, line 47: This line isn't needed, and won't do anything when systemd is the service manager.

PR #635, opened 2023-10-28
podman: workaround ubuntu 22.04 crun bug

Firefishy:
Workaround podman's crun bug in Ubuntu 22.04: https://github.com/containers/podman/issues/9752

The bug seems to happen 1 in 10 in some environments (like: GitHub Actions), so 4 retries drops the odds down to hopefully ~ 0.01%.

### DIFF
--- a/cookbooks/podman/resources/service.rb
+++ b/cookbooks/podman/resources/service.rb
@@ -42,9 +42,12 @@ action :create do
   end
 
   # No action :start here to avoid a start and then immediate :restart, due to subscribe, on first run
+  # FIXME: Ubuntu 22.04 podman/crun bug workaround "retries"
   service new_resource.service do
     action :enable
     subscribes :restart, "systemd_service[#{new_resource.service}]", :immediately
+    retries 4 # Workaround https://github.com/containers/podman/issues/9752
+    retry_delay 5
   end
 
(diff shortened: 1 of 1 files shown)


cookbooks/podman/resources/service.rb
```diff
@@ -42,9 +42,12 @@
   end
 
   # No action :start here to avoid a start and then immediate :restart, due to subscribe, on first run
+  # FIXME: Ubuntu 22.04 podman/crun bug workaround "retries"
   service new_resource.service do
     action :enable
     subscribes :restart, "systemd_service[#{new_resource.service}]", :immediately
+    retries 4 # Workaround https://github.com/containers/podman/issues/9752
+    retry_delay 5
```
tomhughes, line 50: What are we actually retrying here? Not the enable so is it the restart? or the start that is triggered by the next block?

Will a retry actually affect the notified start?

We already have `Restart=on-failure` in the unit anyway so is the concern here one of chef seeing a failure?

Firefishy: It is primarily there as a retry for the `:start` action which occasionally fails due to the crun bug. AFAIK the retry applies to any actions run against the resource.

PR #638, opened 2023-11-17
Add support for using osm2pgsql-replication on render servers

tomhughes:


cookbooks/tile/recipes/default.rb
```diff
@@ -522,77 +522,134 @@
...
-end
 
-service "replicate" do
-  action [:enable, :start]
-  subscribes :restart, "template[/usr/local/bin/replicate]"
-  subscribes :restart, "systemd_service[replicate]"
+  osm2pgsql_arguments.append("--multi-geometry") if node[:tile][:database][:multi_geometry]
+  osm2pgsql_arguments.append("--hstore") if node[:tile][:database][:hstore]
+  osm2pgsql_arguments.append("--tag-transform-script=#{node[:tile][:database][:tag_transform_script]}") if node[:tile][:database][:tag_transform_script]
+
```
pnorman, line 632: Looking at the docs and https://github.com/openstreetmap/operations/issues/987#issuecomment-1816047307 we'll need to specify the --style file.

pnorman (changes requested):
Just the --style addition and it looks good, although I can't test it.

Once this is done, let's stop one of the Australia machines, switch it to osm2pgsql, and restart it

cookbooks/tile/recipes/default.rb
tomhughes: The `--style` does seem to be in the properties table?

```
gis=# select * from osm2pgsql_properties;
       property        |                                value                                 
-----------------------+----------------------------------------------------------------------
 attributes            | false
 db_format             | 2
 flat_node_file        | /store/database/nodes
 import_timestamp      | 2023-10-02T00:00:02Z
 output                | pgsql
 prefix                | planet_osm
 replication_timestamp | 2023-10-02T00:00:02Z
 style                 | /srv/tile.openstreetmap.org/styles/default/openstreetmap-carto.style
 updatable             | true
 version               | 1.9.2
 current_timestamp     | 2023-11-22T14:51:59Z
(11 rows)
```

pnorman: Oh, I guess it's fine then. I guess technically --style is present on the flex backend, although it means something different there.

pnorman (approved):

PR #641, opened 2023-11-28
Add test for chef cookbook

Firefishy:
Add basic chef test to chef service + timer installed and the binary functions.

Added workaround for missing Debian ARM builds by using ARM build from Ubuntu.


tomhughes:
Can we split the debian/arm hack into a separate commit to the test?

PR #645, opened 2023-12-15
Add link to planet.osm Supporting OSM section

hdevine825:


tomhughes:
I believe there is already a plan in the works to do something along these lines but rather more substantial than just a link.

grischard:
@tomhughes that's already done (see ops chat), Harrison is applying the final coat of paint

tomhughes:
Well thanks for given me a chance to review properly before merging.

PR #652, opened 2024-03-19
apache: disable mod_evasive

Firefishy:
mod_evasive is a heavy hammer. At the moment it requires further tuning to avoid false positives. Further experimentation is required to see if a suitable solution can be found.


tomhughes (changes requested):
NACK

PR #654, opened 2024-04-02
Enable use of podman to run tests

tomhughes:
This adds support for using podman to run tests - you just need to run `systemctl --user enable --now podman.socket` to enable the podman socket for your user and that you can run tests as normal.

Note that until https://github.com/test-kitchen/kitchen-dokken/pull/323 is merged you need to set `DOCKER_HOST=unix:///run/user/$(id -u)/podman/podman.sock` in the environment to make it work.


Firefishy:
I tested this on a Mac using `podman machine`... I was only able to get this to work after stripping out the `docker_host_url` lines and just setting `podman = true`

Podman uses an unconventional socket. 
`DOCKER_HOST=unix:///var/folders/kv/sb3s4k3n0zq27rtr2yd_2lc00000gn/T/podman/podman-machine-default-api.sock`
Podman can also set a `/var/run/docker.sock` symlink, but this isn't done by default due to requiring extra permissions.

Firefishy:
Likely complicated by the use of podman machine... But I appears to be rootless

`podman info` reports:

```
host:
  security:
    rootless: true
 ```

tomhughes:
What do you mean by using `podman machine` though? You shouldn't be interacting with podman directly at all, just with kitchen by doing something like `bundle exec kitchen test apache-ubuntu-2204`.

PR #657, opened 2024-04-11
dhcpd: set ignore-client-uids to ensure consistent DHCP IPs

Firefishy:
Without this fix systemd-networkd, PXE, dhclient etc may be given different IPs because of generated client uid mismatches.

### DIFF
--- a/cookbooks/dhcpd/templates/default/dhcpd.conf.erb
+++ b/cookbooks/dhcpd/templates/default/dhcpd.conf.erb
@@ -21,6 +21,9 @@ subnet <%= address.network %> netmask <%= address.netmask %> {
   option PCode "UTC0";
   option TCode "Etc/UTC";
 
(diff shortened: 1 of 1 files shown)


Firefishy:
This is why before https://github.com/openstreetmap/chef/commit/f63f514bf69ed5cfcdb26f9acd6a6209bba3a52a we were getting different IPs for systemd-networkd dhcp and dhclient.

tomhughes:
Our fixed hosts are all locked to a MAC address so the client ID shouldn't any difference - it's more complicated for DHCPv6 for for v4 that should be true.

I think your analysis of https://github.com/openstreetmap/chef/commit/f63f514bf69ed5cfcdb26f9acd6a6209bba3a52a is wrong - my memory is that ifupdown was starting a DHCP client for underlying interfaces in the bond which might not have the same MAC address?

PR #658, opened 2024-04-18
otrs: Add otrs::debian for installing otrs via deb package

Firefishy:
Add chef cookbook `otrs::debian` for installing OTRS/Znuny from debian's otrs2/znuny package.

### DIFF
--- a/cookbooks/otrs/recipes/debian.rb
+++ b/cookbooks/otrs/recipes/debian.rb
@@ -1,0 +1,121 @@
+#
+# Cookbook:: otrs
+# Recipe:: debian
+#
+# Copyright:: 2024, OpenStreetMap Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+include_recipe "accounts"
+include_recipe "apache"
+include_recipe "exim"
+include_recipe "postgresql"
+include_recipe "tools"
+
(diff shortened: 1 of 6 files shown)


cookbooks/otrs/recipes/debian.rb
```diff
@@ -0,0 +1,115 @@
...
+  owner "root"
+  group "root"
+  mode "600"
+  variables :database_name => database_name,
+            :database_user => database_user,
+            :database_password => database_password
+end
+
+apt_package "otrs2" do
+  options "-t #{node[:lsb][:codename]}-backports"
```
tomhughes, line 66: Why do we need this? Is backports not enabled? If it isn't then I think we should enable it instead of doing this or we won't get updates.

tomhughes (commented):
I'm going to need to review this properly when I'm back home, and also do a lot of research on what exactly `dbconfig` is doing as it looks pretty scary to me.

cookbooks/otrs/recipes/debian.rb
Firefishy: Debian backports are masked by default and this is the required option to install from backports.

tomhughes: What does "masked" mean exactly? Just that they won't install new packages from there but will update existing ones?

Firefishy: To install a backport package you have to specify backports as the package source when installing. Debian does this via giving backports a lower pinning priority over the other repos. When a package from backports is installed the pinning priority on that package is changed to allow upgrades. It is detailed here: https://backports.debian.org/Instructions/#index3h2

tomhughes (changes requested):
It's probably worth changing the postgres version in the otrs attributes file to 16 as part of this, so long as we only merge this at the same time as moving the role to the new machine.

cookbooks/otrs/recipes/debian.rb
```diff
@@ -0,0 +1,120 @@
...
+end
+
+# Disable deb otrs2 apache config
+apache_conf "otrs2" do
+  action :disable
+end
+
+# Disable deb otrs2 cron job
+file "/etc/cron.d/otrs2" do
+  action :delete
```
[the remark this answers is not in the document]
Firefishy: Amended.

PR #660, opened 2024-04-22
Allow wordpress themes to be installed from zip

grischard:


cookbooks/wordpress/resources/theme.rb
```diff
@@ -52,6 +52,21 @@
...
+      zip_path = "#{Chef::Config[:file_cache_path]}/#{new_resource.theme}.zip"
+
+      remote_file zip_path do
+        source theme_repository
+        action :create
+      end
+
+      archive_file zip_path do
+        destination theme_directory
+        action :extract
```
tomhughes, line 65: This should be `action :nothing` and then subscribe to changes in the zip with `subscribes :extract, "remote_file[#{zip_path}]` or it will wind up being unpacked every time chef runs.

tomhughes (changes requested):

cookbooks/wordpress/resources/theme.rb
grischard: That's much better indeed.

grischard:
This is only necessary for themes for now (https://github.com/openstreetmap/operations/issues/1036), but could easily be adapted for plugins if necessary.

OK to merge after the tests run?

PR #661, opened 2024-04-23
Chef install morden and varia themes for supporting.o.o

grischard:


cookbooks/civicrm/recipes/default.rb
```diff
@@ -91,6 +91,18 @@
   site "supporting.openstreetmap.org"
 end
 
+wordpress_theme "morden" do
+  theme "morden"
```
tomhughes, line 95: You don't need to repeat the theme name if it's the same as the resource name - it's the name property so is defaulted automatically.

tomhughes (changes requested):

Firefishy:
I am wondering if switching the code to [wp-cli theme install](https://developer.wordpress.org/cli/commands/theme/install/)  might be a better option.

tomhughes:
> I am wondering if switching the code to [wp-cli theme install](https://developer.wordpress.org/cli/commands/theme/install/) might be a better option.

We already discussed that - does that actually do the install itself or does it just call the API on the site and ask it to do the install, which will then have the same permissions issues as using the web interface?

grischard:
@Firefishy this also works, and the wp-cli path runs into handling output, permissions, etc. Unless we can think of concrete advantages to doing this with wp-cli, let's not let perfect be the enemy of good?

PR #663, opened 2024-04-24
systemd-networkd-wait-online waiting for timeout in dokken

Firefishy:
This lets `systemd-networkd` think it is managing the networking in podman / docker and ensures that `systemd-networkd-wait-online` does not wait for timeout on systemd services which rely on networking.

### DIFF
--- a/cookbooks/networking/recipes/default.rb
+++ b/cookbooks/networking/recipes/default.rb
@@ -266,6 +266,16 @@ if node[:networking][:wireguard][:enabled]
   end
 end
 
+# Setup dokken network in systemd-networkd to avoid systemd-networkd-wait-online delay
+template "/etc/systemd/network/dokken.network" do
+  source "dokken.network.erb"
+  owner "root"
+  group "root"
+  mode "644"
+  notifies :run, "execute[networkctl-reload]", :immediately
+  only_if { kitchen? }
+end
+
 notify_group "networkctl-reload"
 
 execute "networkctl-reload" do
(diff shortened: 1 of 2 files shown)


tomhughes:
Are we deliberately enabling wait-online or is something else enabling it? It's not normally needed...

Firefishy:
> Are we deliberately enabling wait-online or is something else enabling it? It's not normally needed...

We don't deliberately enable it, it is part of the dependency for some systemd services. Any service which depends on `systemd-networkd` will trigger `systemd-networkd-wait-online`. eg: systemd-resolved triggers it, but likely others too.

Firefishy:
This shaves for around 90 seconds per test that is using networking: apache2, exim, etc.

tomhughes:
> We don't deliberately enable it, it is part of the dependency for some systemd services. Any service which depends on `systemd-networkd` will trigger `systemd-networkd-wait-online`. eg: systemd-resolved triggers it, but likely others too.

No that's not how it works... There's a discussion at https://systemd.io/NETWORK_ONLINE/ but the important thing is that if a unit is configured to order itself after `network-online.target` then you will need a service like `systemd-networkd-wait-online.service` that can trigger it but the general recommendation is to write services so that don't need the network to be online (partly because nobody can agree what that means) and just order after `network.target` instead so I was interested in what was waiting on `network-online.target` which should itself be the only thing waiting on the networkd service.

tomhughes (approved):
I can't really explain it because I can't find anything in an apache test container that is actually waitng for the network but you're right that it does seem to make a big difference so I guess it's good.

PR #664, opened 2024-04-24
Sparse clone the prometheus-exporters repo

Firefishy:
This sparse clones the prometheus-exporters, while maintaining chef git resource compatibility.

Chef git resource continues as expected.

Conditionally sparse-checkout the exporters and collectors as required.

### DIFF
--- a/cookbooks/prometheus/recipes/default.rb
+++ b/cookbooks/prometheus/recipes/default.rb
@@ -46,6 +46,13 @@ directory "/opt/prometheus" do
   recursive true
 end
 
+execute "git-sparse-clone-prometheus-exporters" do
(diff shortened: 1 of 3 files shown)


tomhughes:
What part of "we need to find a way to solve this properly" was hard to understand?

Firefishy:
> What part of "we need to find a way to solve this properly" was hard to understand?

I am more than happy for a better solution to replace this, until then I hope we can use something like this to shave many minutes off test run times.

tomhughes:
So I just tried this with the apache-ubuntu-2204 test and bizarrely it made no difference to the run time which I do admit I don't really understand.

Firefishy:
Ubuntu 20.04 is too old for the sparse checkout, but Ubuntu 22.04, 24.04 and Debian 12 is faster in my testing. Saving of at least 40 seconds per test which pulls in prometheus. Speed very dependent on clone speed from github, which can vary a lot.

PR #665, opened 2024-04-30
Set dnssec negative trust anchors

Firefishy:
Set DNSSSEC negative trust anchors.

* `geo.openstreetmap.org` because systemd-resolved in Ubuntu 22.04 seems to occasionally get confused if this zone is DNSSEC signed and resolving domains in this zone fail (eg: `nominatim.geo.openstreetamp.org`). Might be related to upstream DNS server features.
* `10.in-addr.arpa` unsigned local zone.
* `internal` used by containers.

### DIFF
--- a/cookbooks/networking/recipes/default.rb
+++ b/cookbooks/networking/recipes/default.rb
@@ -307,6 +307,20 @@ service "systemd-resolved" do
   action [:enable, :start]
 end
 
+directory "/etc/dnssec-trust-anchors.d" do
+  owner "root"
+  group "root"
+  mode "755"
+end
+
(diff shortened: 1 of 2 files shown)


Firefishy:
Added in the default NTAs.

Firefishy:
I am closing this PR. It appears the issue we had a few days back is a bit more complex. On gateway instances we have `bind9 / named` running a DNS server. When DNS stopped working restarting JUST `named` on the gateway fixed the issue.

Somehow the DNSSEC was failing for `openstreetmap.org` and not the `geo.openstreetmap.org` zone. There belief is something was likely broken with `named` cache or similar.

PR #667, opened 2024-05-02
web: add openstreetmap.com to certificate

Firefishy:
`openstreetmap.COM` is a commonly used incorrect domain for us.

The domain displays an HTTPS certificate error. Some users may still have a redirect cached and will not see the error (Redirect cached until 2038)

This PR adds the certificate so that the `openstreetmap.COM` will be correctly redirected without a potential certificate issue.


Firefishy:
@tomhughes Added you as reviewer to see if this change seems reasonable.

tomhughes (approved):

PR #673, opened 2024-05-28
apache: use 429 error for mod-evasive

Firefishy:
Currently mod-evasive uses `403 - Forbidden`, instead use `429 - Too Many Requests` instead to clearly identify the responses when mod-evasive is triggered.


Firefishy:
This works because upstream accepted this patch: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=930724

mmd-osm:
As a heads up, we're facing quite some issues related to mod_evasive. It would be great if you could review my analysis in https://github.com/facebook/Rapid/issues/1424 ... we'll create any operations issue later.

tomhughes (approved):

PR #674, opened 2024-06-12
Add prometheus alert for too many parallel connections to nominatim

lonvia:
My first attempt at prometheus alert rule writing. Not sure I got it right.

What I'm trying to do is to trigger an alert when the number of connections in writing in [Nominatim's 'Active Connections by State' graph](https://prometheus.openstreetmap.org/d/6qvCSGtnz/nominatim) goes over 2.5k. This intentionally works on the sum of all Nominatim servers because it'd need different limits for the different servers otherwise.

### DIFF
--- a/cookbooks/prometheus/templates/default/alert_rules.yml.erb
+++ b/cookbooks/prometheus/templates/default/alert_rules.yml.erb
@@ -488,6 +488,11 @@ groups:
           alertgroup: nominatim
         annotations:
           delay: "{{ $value | humanizeDuration }}"
+      - alert: nominatim connections
+        expr: sum(nginx_connections_writing and on (instance) chef_role{name="nominatim"}) > 2500
+        for: 15m
+        labels:
+          alertgroup: nominatim
   - name: overpass
     rules:
       - alert: overpass osm database age


cookbooks/prometheus/templates/default/alert_rules.yml.erb
```diff
@@ -488,6 +488,11 @@ groups:
           alertgroup: nominatim
         annotations:
           delay: "{{ $value | humanizeDuration }}"
+      - alert: nominatim connections
+        expr: sum(nginx_connections_writing{instance=~chef_role{name="nominatim"}}) > 2500
```
tomhughes, line 492: That's not going to work, it's not even a valid expression. I think what you want is:

```
sum(nginx_connections_writing and on (instance) chef_role{name="nominatim"}) > 2500
```

The easy way to test expression is to past then into [Explore](https://prometheus.openstreetmap.org/explore?schemaVersion=1&panes=%7B%224bo%22:%7B%22datasource%22:%22oCW3xcdMz%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22sum%28nginx_connections_writing%20and%20on%20%28instance%29%20chef_role%7Bname%3D%5C%22nominatim%5C%22%7D%29%20%3E%202500%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22oCW3xcdMz%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1).

tomhughes (changes requested):

cookbooks/prometheus/templates/default/alert_rules.yml.erb
lonvia: Even comes with an expression builder. Very useful.

I've fix the expression.

PR #679, opened 2024-07-10
CiviCRM

datendelphin:
- make changes persistent from last tricky civicrm update
- update osmf verify contributor plugin
- write civicrm config to new location, old one is deprecated since a few years

Please check the config file currently on the production server (there is one already in the new location) before running chef. If there are differences, we might need to reflect them in chef as well.


datendelphin:
Thanks for supplying the diff. I checked it, and it is mostly updates to the template configuration file. However, the file that lies currently in the new location does not contain some keys. So probably the old location was still in use.

You can merge it now

PR #682, opened 2024-07-22
apache: only show major/minor apache version in response header

Firefishy:
Only show show `Server: Apache/2.4` instead of full `Server: Apache/2.4.54 (Ubuntu)` response header.

The full version is unnecessary level detail and may cause false positive security reports.


tomhughes:
Well they're not "false positive" they're just idiotic.

PR #684, opened 2024-08-05
Restrict frontend access to CDNs, monitoring and admins

grischard:


grischard:
That's for tiles from three years ago, not the frontends

PR #685, opened 2024-08-07
Add docker hub login to avoid hub IP limits

Firefishy:
Workaround anonymous limits of Docker Hub registry by having the GHA login.


tomhughes:
What login is that using though?

Firefishy:
> What login is that using though?

It is using a dedicated Docker Hub personal access token from my Docker Hub account with only "Public Repo Read-only" permission.

Firefishy:
I've now changed this PR to not run the login step when running from a PR: `if: github.repository == 'openstreetmap/chef' && github.event_name != 'pull_request'`

The PR is still useful because it will still trigger on non-forked branch pushes eg: master.

PR #688, opened 2024-08-23
Add roles for CloudFerro servers

pnorman:


roles/cloudferro-waw3-2.rb
```diff
@@ -0,0 +1,7 @@
+name "cloudferro-waw3-2"
+description "Role applied to all servers at CloudFerro WAW3-2"
+
```
tomhughes, line 3: This should probably set a location? Something like:

```ruby
default_attributes(
  :location => "Warsaw"
)
```

tomhughes (approved):
Looks good to me now.

PR #691, opened 2024-09-03
Install system pyotp

pnorman:
This avoids failures on Debian systems.

### DIFF
--- a/cookbooks/tile/recipes/default.rb
+++ b/cookbooks/tile/recipes/default.rb
@@ -159,13 +159,10 @@ end
 package %w[
   python3-cairo
   python3-mapnik
+  python3-pyotp
   python3-pyproj
   python3-setuptools
 ]
-
-python_package "pyotp" do
-  python_version "3"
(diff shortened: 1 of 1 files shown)


Firefishy:
Could be replaced with package: https://packages.debian.org/bookworm/python3-pyotp

tomhughes:
Well yes, that's exactly what this PR does?

cookbooks/tile/recipes/default.rb
```diff
@@ -161,12 +161,9 @@
   python3-mapnik
   python3-pyproj
   python3-setuptools
+  python3-pyotp
```
tomhughes, line 164: Can we move this up so it's in alphabetical order like the others?

tomhughes (changes requested):

cookbooks/tile/recipes/default.rb
pnorman: done

PR #694, opened 2024-09-12
apt: Source kernel and firmware from backports

Firefishy:
We've had performance issues with the default 6.1.x kernel in debian. The backports kernel fixes the issue.


Firefishy:
Manually tested on faffy and ironbelly with no observed ill effects.

Firefishy:
Overlap of work with @tomhughes

PR #696, opened 2024-09-12
Add cmok server

pnorman:
This is for the new vector tile server that will be running at CloudFerro

### DIFF
--- a/roles/cmok.rb
+++ b/roles/cmok.rb
@@ -1,0 +1,38 @@
+name "cmok"
+description "Master role applied to cmok"
+
+default_attributes(
+  :networking => {
+    :intefaces => {
+      :external => {
+        :interface => "ens3",
+        :role => :external,
+        :inet => {
+          :prefix => "24",
+          :gateway => "10.0.0.1",
+          :public_address => "64.225.143.127"
+        }
+      }
+    }
+  },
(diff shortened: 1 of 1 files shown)


roles/cmok.rb
```diff
@@ -0,0 +1,38 @@
...
+description "Master role applied to cmok"
+
+default_attributes(
+  :networking => {
+    :intefaces => {
+      :external => {
+        :interface => "ens3",
+        :role => :external,
+        :inet => {
+          :prefix => "24",
```
tomhughes, line 11: We don't know yet what the internal address will be?

pnorman: No. It will be on the same network as wawel, but the 10.x IPs are not statically assigned.

tomhughes: Well wawel's is being statically assigned now... I asked about whether it needed to be DHCP and was told now. If it needs to be DHCP then that will need some work in the cookbooks.

Firefishy: OpenStack neuron assigns the internal IP for the life of the VM (or more specially the interface). There is no need for DHCP. If we change the interface or the emulated type (not that we would, virt-nic is best) then we would briefly need to switch back to DHCP.

PR #721, opened 2024-12-24
vectortile: Use flex expiry

pnorman:
This uses the new flex expiry code in spirit/shortbread to reduce the number of tiles that need to be re-rendered each update.

### DIFF
--- a/cookbooks/vectortile/templates/default/import-planet.erb
+++ b/cookbooks/vectortile/templates/default/import-planet.erb
@@ -7,7 +7,7 @@
 
 set -e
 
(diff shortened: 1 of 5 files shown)


cookbooks/vectortile/templates/default/import-planet.erb
```diff
@@ -7,7 +7,7 @@
 
 set -e
 
-export LUA_PATH='/srv/vector.openstreetmap.org/osm2pgsql-themepark/lua/?.lua;/srv/vector.openstreetmap.org/spirit/?.lua;;'
```
pnorman, line 10: This bit is no longer necessary as a change resulting from lua include path changes made in spirit to allow expiry to work

PR #723, opened 2025-01-08
Update YAML.safe_load arguments in replicate-changesets

AntonKhorev:
I wanted to check how replicate-changesets works with https://github.com/openstreetmap/openstreetmap-website/pull/5309 and it didn't run for a number of reasons. First, `YAML.safe_load` arguments are outdated since https://github.com/ruby/psych/commit/0767227051dbddf1f949eef512c174deabf22891 and an update like https://github.com/openstreetmap/chef/commit/8061c292a2490f50f1e84a5bccde1b45e5d3d511 is needed.


tomhughes:
Looks good to me, thanks.

PR #724, opened 2025-01-08
Use empty hash fallback when loading state file in replicate-changesets

AntonKhorev:
If I use readme at https://github.com/zerebubuth/openstreetmap-changeset-replication as a guide, it says "a state file, which can be empty the first time it's used". But that's not going to work because an empty file is read as `nil`, then the code tries to access elements of that `nil` and crashes.

Here I change an empty state_file to be read as an empty hash.


AntonKhorev:
I didn't get the code fully working yet, there are utc vs local timezone issues. But looks like it won't start from the first changeset.

tomhughes:
I guess this can't do any harm anyway.

AntonKhorev:
It starts replicating from current time.

PR #725, opened 2025-01-10
Terminate fdirsync loop when "." is reached in replicate-changesets

AntonKhorev:
Looking at the readme in https://github.com/zerebubuth/openstreetmap-changeset-replication you can see this line in the example config:

```
data_dir: replication/
```

Looks like you can specify a relative path as `data_dir`. But actually you can't, because the loop I modify in this PR will never terminate. The server config avoids this issue by using an absolute path:

https://github.com/openstreetmap/chef/blob/e4330da209a809e67d03ff9e2b0a7f4bcac55863/cookbooks/planet/templates/default/changesets.conf.erb#L3


tomhughes:
Looks good to me, thanks.

PR #726, opened 2025-01-10
Select db dates at utc in replicate-changesets

AntonKhorev:
- `select now()` gives you time with timezone
- `select created_at` etc give you time without timezone because we don't store timezones
- `Time.parse()` on time without timezone parses it as a local time
- comparisons between times with/without timezones are broken both in ruby and in db queries

To fix all this I'm adding `at time zone 'utc'` to every time value from db that doesn't have a timezone.

### DIFF
--- a/cookbooks/planet/files/default/replication-bin/replicate-changesets
+++ b/cookbooks/planet/files/default/replication-bin/replicate-changesets
@@ -96,7 +96,7 @@ class ChangesetBuilder
 
   def add_comments(xml, cs)
     # grab the visible changeset comments as well
-    res = @conn.exec("select cc.author_id, u.display_name as author, cc.body, cc.created_at from changeset_comments cc join users u on cc.author_id=u.id where cc.changeset_id=#{cs.id} and cc.visible order by cc.created_at asc")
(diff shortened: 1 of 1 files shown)


tomhughes:
Presumably this is only a problem when running a machine in a non-UTC timezone?

AntonKhorev:
Probably, but there was one `at time zone 'utc'` before my changes, added in https://github.com/zerebubuth/openstreetmap-changeset-replication/commit/6c2aee58d0656dba2a7a4cb8c16a9b4771f4f32a. Why would that be needed if the goal was to make it run only in utc?

tomhughes:
My point was that as far as I know there are no problems in production, but that machine like all our machines runs in UTC all the time.

tomhughes:
I guess this is fine then, thanks.

I do wonder if all these changes should be going in the upstream repo first? or are we treating that as dead now?

AntonKhorev:
Even @zerebubuth changed the script here https://github.com/openstreetmap/chef/commit/60735f534585baea7c17d0320c22bb56f19ddabf without updating https://github.com/zerebubuth/openstreetmap-changeset-replication

PR #731, opened 2025-01-30
enable webp support for imagery layers

andrewharvey:
We have WebP layers in the editor-layer-index now https://github.com/osmlab/editor-layer-index/issues/1890 and it would be nice to access the OSM hosted imagery layers via WebP for lower bandwidth with transparency support.

I'm unsure about any CPU load implications.


Firefishy:
Cheers!

Firefishy:
Just a note, some layers use titiler as the backend, which supports webp. Others like (AGRI) use mapserver which currently does not support webp.

andrewharvey:
Thanks, I saw that and from what I could tell the format is not passed through to mapserver anyway so has no affect there.

https://github.com/andrewharvey/chef/blob/d6d65b2e627e46df6bdaef4a6135e598c64038ae/cookbooks/imagery/templates/default/nginx_imagery_layer_fragment.conf.erb#L36

PR #733, opened 2025-02-01
add au_act_aerial mapserver imagery

andrewharvey:
We have [permission to use](https://wiki.openstreetmap.org/wiki/File:ACT_Government_20180314_Permission_to_Incorporate_CCBY_data_into_OSM.PDF) ACT CC BY imagery in OSM and have a [few layers in the ELI](https://github.com/osmlab/editor-layer-index/tree/gh-pages/sources/oceania/au/act), however recently these have only been published as WMTS with EPSG:7855.

This works in JOSM, but unfortunately WMTS won't (unlike WMS) accept a request to serve images in Web Mercator and iD doesn't support reprojecting image tiles from EPSG:7855 to WebMercator.

However since you're running a MapServer service already, perhaps we can include the configuration needed to proxy these layers from EPSG:7855 tiles to Web Mercator tiles?

I've tested a version of this mapserver .map file locally and it seems to be working, but I haven't tested the cookbooks (sorry no experience with that).

cc @Firefishy

### DIFF
--- a/cookbooks/imagery/recipes/au_act_aerial.rb
+++ b/cookbooks/imagery/recipes/au_act_aerial.rb
@@ -1,0 +1,67 @@
+#
+# Cookbook:: imagery
+# Recipe:: au_act_aerial
+#
+# Copyright:: 2025, OpenStreetMap Foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


andrewharvey:
More context at https://github.com/osmlab/editor-layer-index/issues/2340.

I've also emailed the department asking about other imagery sources since they used to publish bi-annual going back a while.

cc @FrakGart @davidhicks

Firefishy:
Looks good. Merged.

PS: We have some simple tests that can also run locally, see https://github.com/openstreetmap/chef/blob/master/CONTRIBUTING.md

Firefishy:
Up and running: https://act-imagery.openstreetmap.org/

andrewharvey:
Thanks!

PR #735, opened 2025-02-03
Add copyright/attribution message to downloaded map images

mmd-osm:
This PR adds OSM attribution to the map export, both for vector and raster formats. Besides, this PR adds webp image format.

Image

I introduced a new dependency on python3-pil package to enable the png->{webp,jpeg} conversion.

Fixes https://github.com/openstreetmap/openstreetmap-website/issues/551

### DIFF
--- a/cookbooks/tile/recipes/default.rb
+++ b/cookbooks/tile/recipes/default.rb
@@ -159,6 +159,7 @@ end
 package %w[
   python3-cairo
   python3-mapnik
+  python3-pil
   python3-pyotp
   python3-pyproj
   python3-setuptools
--- a/cookbooks/tile/templates/default/export.erb
+++ b/cookbooks/tile/templates/default/export.erb
@@ -5,6 +5,7 @@ import cairo
 import cgi
 import http.cookies
 import mapnik
+import io
 import os
 import pyotp
 import pyproj
@@ -13,6 +14,7 @@ import shutil
 import signal
 import sys
 import tempfile
+from PIL import Image
 
 # Limit maximum CPU time
 # The Postscript output format can sometimes take hours
@@ -40,6 +42,10 @@ def output_file(file):
 def file_size(file):
   return os.fstat(file.fileno()).st_size
 
+# Routine to retrieve BytesIO payload length
+def bytesio_size(bio):
+  return bio.getbuffer().nbytes
+
 # Routine to report an error
 def output_error(message, status = "400 Bad Request"):
   print("Status: %s" % status)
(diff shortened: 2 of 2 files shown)


cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +55,115 @@ def output_error(message, status = "400 Bad Request"):
   print("</body>")
   print("</html>")
 
+# Add a copyright notice for raster formats (PNG, JPEG, WEBP)
+def add_copyright_notice_raster(image, map_width, map_height, format):
+  #  Create a temporary file for the PNG image
+  with tempfile.NamedTemporaryFile(suffix=".png") as png_tempfile:
+    # Save the Mapnik image into the temporary file as PNG
+    png = image.tostring("png")
+    png_tempfile.write(png)
```
tomhughes, line 64: Can we not just do `io.BytesIO(png)` to create an in-memory file object that can be passed to the surface constructor without involving a temporary file?

cookbooks/tile/templates/default/export.erb
```diff
@@ -13,6 +13,7 @@ import shutil
 import signal
 import sys
 import tempfile
+from PIL import Image
```
tomhughes, line 16: Do we need to change the recipe to ensure `PIL` is installed?

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +55,115 @@ def output_error(message, status = "400 Bad Request"):
...
+
+    # Load the PNG data from the temporary file into a Cairo ImageSurface
+    surface = cairo.ImageSurface.create_from_png(png_tempfile.name)
+
+  add_copyright_notice_vector(surface, map_width, map_height)
+
+  # Create a temporary file for the output image
+  with tempfile.NamedTemporaryFile(suffix="." + format) as img_tempfile:
+    # Convert the Cairo surface to PNG and save it to the temporary file
+    surface.write_to_png(img_tempfile.name)
```
tomhughes, line 74: Same on the way out, this can write to a `BytesIO` object?

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +55,115 @@ def output_error(message, status = "400 Bad Request"):
...
+
+    if format == "png":
+      img_tempfile.seek(0)
+      return img_tempfile.read()
+    else:
+      # Open the output PNG image for conversion to other formats
+      img = Image.open(img_tempfile.name)
```
tomhughes, line 81: I think `Image.frombytes` could be used here to ingest the backing data from the in-memory file?

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +55,115 @@ def output_error(message, status = "400 Bad Request"):
...
+      with tempfile.NamedTemporaryFile() as f:
+        img.save(f, format=format, quality=100)
```
tomhughes, line 83: Then write back to a `BytesIO` stream again?

mmd-osm:
So regarding the temporary file vs. BytesIO topic. I looked at both options and thought it might be safer to store intermediate results on disk, rather than trying to juggle everything in memory. I also noticed the "resource.setrlimit" at the beginning and thought that memory usage is something to watch out for.

Are temporary files too much of a concern?

cookbooks/tile/templates/default/export.erb
mmd-osm: Probably yes, but I can't tell for sure if PIL hasn't been pulled in by some other dependency already.

tomhughes: Well if it's required we should add it to the install list rather than relying on something else pulling it in accidentally - even if that works today it might not in the future.

mmd-osm: If https://github.com/openstreetmap/chef/blob/master/cookbooks/tile/recipes/default.rb#L102-L105 is the right place, I could of course add it there.

tomhughes:
The memory limit is mostly just a backstop against people doing stupid things, especially with the vector output formats as I recall.

Temporary files generally make me nervous because you need to be sure they will always get cleaned up properly, especially if like here you're not immediately deleting them, which you can't if you need to be able to pass them to functions by name. Though all the functions here could actually take a handle instead I think.

cookbooks/tile/templates/default/export.erb
tomhughes: https://github.com/openstreetmap/chef/blob/24b4995f48385560ec2a51672f1113f02dadd4cd/cookbooks/tile/recipes/default.rb#L159-L165 is probably better.

cookbooks/tile/templates/default/export.erb
mmd-osm: I’m switched to BytesIO instead of temporary files now. I did a short smoke test with png and webp output, and that looked ok. I hope I didn’t miss anything…

cookbooks/tile/templates/default/export.erb
mmd-osm: Added now

pnorman:
> memory usage is something to watch out for.

I suspect this is because Mapnik can get memory crazy with some output formats. The memory limit is 4GB, so I don't see that storing the rasterized output in-memory is an issue.

PDF and PS are terrible to use with Mapnik and likely to break stuff in other ways, so I wouldn't worry about those.

cookbooks/tile/recipes/default.rb
```diff
@@ -162,6 +162,7 @@
   python3-pyotp
   python3-pyproj
   python3-setuptools
+  python3-pil
```
tomhughes, line 165: Can we keep this list alphabetically ordered please?

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +56,112 @@ def output_error(message, status = "400 Bad Request"):
...
+
+  add_copyright_notice_vector(surface, map_width, map_height)
+
+  # Convert the Cairo surface to PNG in a BytesIO object
+  output_io = io.BytesIO()
+  surface.write_to_png(output_io)
+
+  if format == "png":
+    output_io.seek(0)
+    return output_io.read()
```
tomhughes, line 76: Why are we reading in the data rather than just returning the handle and letting the called use `output_file` on it? If you do that you don't need the seek either as `output_file` already does that.

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +56,112 @@ def output_error(message, status = "400 Bad Request"):
...
+  context.set_source_rgb(0, 0, 0)  # Black color for the text
+  context.move_to(x_position - x_margin / 2, y_position + y_margin)
+  context.show_text(text)
+
+# Render and output map for raster formats (PNG, JPEG, WEBP)
+def render_and_output_image(map, format, attribution):
+  image = mapnik.Image(map.width, map.height)
+  mapnik.render(map, image)
+
+  if attribution == 'X':
```
tomhughes, line 123: Why are we making the attribution optional?

tomhughes (changes requested):

cookbooks/tile/templates/default/export.erb
AntonKhorev: Let's suppose I want to put a static map image on my website. If the attribution is optional, I might export the image without an attribution, put it on my site and add a link to osm under it. If the attribution is not optional, I'll see that osm is already credited and I won't add a link.

cookbooks/tile/templates/default/export.erb
mmd-osm: The current raster code doesn’t use output_file either (but sys.stdout.buffer.write(image)). I wanted to keep the current code as is. Using output_code would require some attribution dependent logic in line 135. I’m not sure this is better.

cookbooks/tile/templates/default/export.erb
mmd-osm: That’s what the original issue https://github.com/openstreetmap/openstreetmap-website/issues/551 was proposing. Maybe @simonpoole still remembers…

simonpoole: At the time my thoughts were that somebody putting the image on a website has ample ways of providing attribution, but somebody that was just printing it out would have to jump through hoops to add it.

But given that the audience is likely getting more unsophisticated by the day it is probably OK to simply always add it.

mmd-osm: At least in https://github.com/openstreetmap/openstreetmap-website/pull/5607 it’s enabled by default and users need to explicitly disable it.

cookbooks/tile/recipes/default.rb
mmd-osm: done

cookbooks/tile/templates/default/export.erb
mmd-osm: All attribution specific code is gone, the attribution will always be added now.

cookbooks/tile/templates/default/export.erb
mmd-osm: I've created a new function `output_image(file)` to avoid some of the code duplication.  `output_file` doesn't seem to work with BytesIO, and` len(image)`  in lines 128+130+132 wouldn't work with an ImageIO object either.

Maybe I'm misunderstanding what exactly you're suggesting. If you have some specific changes in mind, it's probably more efficient if you could propose them as code.

mmd-osm: I have changed the code accordingly and also added a new function bytesio_size for the payload length.

cookbooks/tile/templates/default/export.erb
```diff
@@ -54,6 +60,105 @@ def output_error(message, status = "400 Bad Request"):
...
+  output_io = io.BytesIO()
+  surface.write_to_png(output_io)
+
+  if format == "png":
+    return output_io
+  else:
+    # Open the output PNG image for conversion to other formats
+    img = Image.open(output_io)
+    img_io = io.BytesIO()
+    img.save(img_io, format=format, quality=100)
```
tomhughes, line 84: I don't think we want `quality=100` here... It seems like a really bad idea for [JPEG](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#jpeg-saving) and for [webp](https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#webp-saving) is there a good reason not to go with the default? A better question for webp is whether to do lossy or lossless I suspect?

tomhughes (changes requested):

cookbooks/tile/templates/default/export.erb
mmd-osm: I guess we can go with the default values. I will remove the quality parameter.

tomhughes (approved):
Thanks. This looks good to me now.

PR #736, opened 2025-02-07
Allow podman to run kitchen tests

apenney:
Due to differences in how podman and docker treat volumes, we have to hack in a workaround for podman to start the chef container up before it runs a suite.  Otherwise you get a failure to find the chef-client binary

https://github.com/test-kitchen/kitchen-dokken/issues/255 has more information on the cause

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -62,6 +62,18 @@ platforms:
         - RUN /usr/bin/apt-get install -y eatmydata
         - RUN echo /usr/lib/$(uname -m)-linux-gnu/libeatmydata.so >>/etc/ld.so.preload
 
+# When using podman we have to manually start chef to workaround a volume issue
+lifecycle:
+  pre_create: |
+    if command -v podman >/dev/null 2>&1; then
+      podman create --name chef-latest --replace docker.io/chef/chef:latest sh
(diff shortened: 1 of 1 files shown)


tomhughes:
The `prune -f` is a bit worrying as that will remove all unused volumes.

Oddly the one volume it didn't remove for me is the one in question, because it was still being used by the `chef-latest` container. The documentation says that `-f` will remove the container as well in that case but it didn't for me and I had to remove the container manually before purge would remove the volume.

tomhughes:
Possibly `podman container rm -v chef-latest` would be better?

tomhughes:
I think we need a `-i` on that as well in case the container doesn't exist...

apenney:
After the world's slowest followup, this should be in better shape now:

* It works properly on the first run (thanks to the || true if a container never existed)
* Cleans up after itself

tomhughes:
I switched it to use `-i` instead of `|| true` as it seems slightly safer in terms of not suppressing all the errors but other than that it all looks good so I've merged it. Thanks for the work.

apenney:
@tomhughes Oh nice, I had no idea -i even existed.. I have other places this'll be handy in, thanks!

PR #744, opened 2025-02-25
Ensure certbot does not create new cert if domains change

Firefishy:
Before if the domains for a certificate were reduced a new certificate was created and counter suffix added to the cert-name

eg:  ```/srv/acme.openstreetmap.org/config/live/wiki.openstreetmap.org-0001/```

The upload script would also use the wrong certificate (without counter suffix).

The fix is to explicitly set the `--cert-name`, allow domain changes with `--renew-with-new-domains` and worthwile to re-use the existing hook `/srv/acme.openstreetmap.org/bin/renew-hook` for uploads.


cookbooks/letsencrypt/templates/default/request.erb
```diff
@@ -2,6 +2,13 @@
 
 # DO NOT EDIT - This file is being maintained by Chef
 
+if [ "$(id -un)" != "letsencrypt" ] || [ "$(id -gn)" != "letsencrypt" ]; then
+  echo "Error: must be running as user/group 'letsencrypt'"
+  exit 1
+fi
```
tomhughes, line 8: I can guess why you've added this but it doesn't actually have anything to do with the PR as such...

cookbooks/letsencrypt/templates/default/request.erb
```diff
@@ -2,6 +2,13 @@
...
+
+cd /srv/acme.openstreetmap.org
```
tomhughes, line 10: Why do you think this is needed? Does this script not work regardless of what directory you're in? It uses absolute paths for everything?

tomhughes (commented):
We should probably rename the `renew-hook` script to `deploy-hook` which is the modern name, and if we specify it here when creating the certificate then we can drop it from the `renew` script as it will be remembered in the renewal configuration for the certificate.

cookbooks/letsencrypt/templates/default/request.erb
Firefishy: I've removed this for now. It was paranoid check if being run manually. If manually run as root it can screwup permissions for later automatic runs.

Firefishy:
> We should probably rename the `renew-hook` script to `deploy-hook` which is the modern name, and if we specify it here when creating the certificate then we can drop it from the `renew` script as it will be remembered in the renewal configuration for the certificate.

I have renamed to `deploy-hook`, but the `renew` script calls `--renew-hook /srv/acme.openstreetmap.org/bin/deploy-hook` because existing certificates will not have the correct renew hook.

cookbooks/letsencrypt/templates/default/request.erb
tomhughes: Actually they do have it because it gets saved in the config on first renewal.

cookbooks/letsencrypt/files/default/bin/renew
```diff
@@ -7,4 +7,4 @@ cd /srv/acme.openstreetmap.org
     --config-dir /srv/acme.openstreetmap.org/config \
     --work-dir /srv/acme.openstreetmap.org/work \
     --logs-dir /srv/acme.openstreetmap.org/logs \
-    --renew-hook /srv/acme.openstreetmap.org/bin/renew-hook
+    --renew-hook /srv/acme.openstreetmap.org/bin/deploy-hook
```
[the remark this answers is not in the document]
Firefishy: Changed to `renew` to use `--deploy-hook`

PR #746, opened 2025-02-28
letsencrypt: add retry logic if cert request fails

Firefishy:
Add logic to retry certificate requests if the request failed on initial chef run.

### DIFF
--- a/cookbooks/letsencrypt/recipes/default.rb
+++ b/cookbooks/letsencrypt/recipes/default.rb
@@ -84,6 +84,12 @@ directory "/srv/acme.openstreetmap.org/logs" do
   mode "700"
 end
 
+directory "/srv/acme.openstreetmap.org/failures" do
+  owner "letsencrypt"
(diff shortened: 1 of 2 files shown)


tomhughes:
This is playing with fire - the risk is that if there is something fundamentally wrong it just keeps retrying until it hits the rate limit.

Better to let a human deal with any failures I think.

Firefishy:
> This is playing with fire - the risk is that if there is something fundamentally wrong it just keeps retrying until it hits the rate limit.
> 

`Up to 50 certificates can be issued per registered domain every 7 days.`

I guess we could hit this one on a repeating failure. Hmmm...

tomhughes:
What exactly is the problem you're trying to solve here? I don't recall ever encountering failures that weren't caused by my screwing something up that needs to be fixed?

Sometimes renewals hit some network glitch but that's why they start trying weeks before needed and they just renew on the next run.

PR #759, opened 2025-05-05
vectortile: Rerender lowzoom tiles daily

pnorman:


cookbooks/vectortile/recipes/default.rb
```diff
@@ -292,6 +292,40 @@
...
+  source "render-lowzoom.erb"
+  owner "root"
+  group "root"
+  mode "755"
+  variables :tilekiln_bin => "#{tilekiln_directory}/bin/tilekiln", :source_database => "spirit", :storage_database => "tiles", :config_path => shortbread_config, :min_zoom => 0, :max_zoom => node[:vectortile][:rerender][:lowzoom][:maxzoom]
+end
+
+systemd_service "render-lowzoom" do
+  description "Render low zoom tiles"
+  condition_path_exists_glob "!/run/render-lowzoom-*"
```
tomhughes, line 305: What is this meant to protect against? If it's protecting against two instances of the service running at once then I don't think it's needed.

pnorman: That's what I intended to protect against. I got it from the existing raster tile rerender job

tomhughes: The raster one is not protecting against itself though - that is `render-lowzoom` which runs on a timer being blocked if `update-lowzoom` which chef starts when the style changes is running.

There's no need for a service to protect against itself because a service can't be running more than once - if the timer asks to start the service and the service is already running then nothing will happen.

pnorman: 👍 Will fix

tomhughes (changes requested):
Can you squash that fixup commit into the original commit please...

PR #769, opened 2025-05-26
Set PGAPPNAME for cgimap

mmd-osm:
In https://en.osm.town/@pnorman/114572540276425975 , @pnorman has recommended  to set the PostgreSQL application_name.

Setting PGAPPNAME as environment variable should achieve this goal without any code changes (confirmed in local test with openstreetmap-cgimap binary) . The change in this PR is based on my understanding how cgimap.rb works. I haven't tested if the ruby script as such.

Maybe this is useful for prometheus to break down the number of connections by client application.


tomhughes:
Is there a reason for not setting this upstream in the code? I mean this might be useful as a temporary measure but long term it seems like something that would be good for cgimap to do by default?

PR #772, opened 2025-06-05
Switch backups to using zstd

Firefishy:
Switch backups to using zstd over gzip.

Use all cores to allow fast backups.

Example: wiki backup is ~75% faster to create and smaller.

### DIFF
--- a/cookbooks/blog/templates/default/backup-staging.cron.erb
+++ b/cookbooks/blog/templates/default/backup-staging.cron.erb
@@ -2,20 +2,21 @@
 
 # DO NOT EDIT - This file is being maintained by Chef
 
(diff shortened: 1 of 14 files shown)


tomhughes (commented):
Why are we setting the compression level and thread count twice, once via the environment and once on the command line?

Then again I don't know why everything is setting `RSYNC_SSH` when we're using the native rsync transport ;-)

Firefishy:
> Why are we setting the compression level and thread count twice, once via the environment and once on the command line?
> 
> Then again I don't know why everything is setting `RSYNC_SSH` when we're using the native rsync transport ;-)

I'll fix both.

PR #773, opened 2025-06-11
exim: Increase message_linelength_limit to avoid dropping emails

Firefishy:
Increase the exim max SMTP message line length so as not to reject some badly formatted emails.

Currently a few `cron.daily` emails are rejected outbound due to line length. (eg: fafnir)

The issue seems common enough that the workaround has been templated:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=828801
https://serverfault.com/questions/842176/exim-limiting-number-of-emails-and-cc/881197#881197

### DIFF
--- a/cookbooks/exim/templates/default/exim4.conf.erb
+++ b/cookbooks/exim/templates/default/exim4.conf.erb
@@ -638,7 +638,7 @@ mailman:
   local_part_suffix = -bounces : -bounces+* : \
                       -confirm+* : -join : -leave : \
                       -subscribe : -unsubscribe : \
-                      -owner : -request : -admin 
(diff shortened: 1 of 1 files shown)


Firefishy:
> I can understand that cron doesn't try and wrap lines but what exactly is going on in those jobs that causes lines over 998 characters long in the first place?

I've been unable to find which script is causing the `cron.daily` email failure. Example ID: `E1uOrcq-00000009naY-2vaN@fafnir.openstreetmap.org`

Yes, while true this violates the SMTP RFC, I don't think it causes any real issues.

tomhughes (approved):
I guess we'll find out what's causing the long line once it goes through.

PR #783, opened 2025-07-19
Map export: replace cgi by urllib

mmd-osm:
+ added exception handler

### DIFF
--- a/cookbooks/tile/templates/default/export.erb
+++ b/cookbooks/tile/templates/default/export.erb
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 
 import cairo
-import cgi
 import http.cookies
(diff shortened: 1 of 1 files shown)


cookbooks/tile/templates/default/export.erb
```diff
@@ -180,7 +181,7 @@ if 'HTTP_REFERER' not in os.environ:
 if '_osm_totp_token' in cookies:
   token = cookies['_osm_totp_token'].value
 elif os.environ['REQUEST_METHOD'] != 'OPTIONS':
-  token = form.getfirst("token")
+  token = form.get("token", [None])[0]
```
tomhughes, line 184: We can presumably just make this a straight `else` now and get rid of the options check?

PR #784, opened 2025-07-21
Add debian 13 to tests

Firefishy:
Add debian-13 to test suite.


Firefishy:
Blocked for now on https://github.com/fullstaq-ruby/server-edition/pull/173

PR #795, opened 2025-08-28
nft: Allow select ICMP types

Firefishy:
Allow additional ICMP message types. We occasionally incorrectly drop some of these.


Firefishy:
> Are we really incorrectly dropping them? They will likely already get through if they're related to an existing tracked connection.

Ah! My mistake in understand. I didn't realise the accept related would allow the ICMP

PR #811, opened 2025-11-26
feat: Add flox for local development

apenney:
[I created this because I use flox for everything else and I like having isolated environments when working on stuff.  It's a great tool, based on nix, and is open source so hopefully this is acceptable!]

Flox (flox.dev) is a nix based tool for building local developer environments.  This work adds a flox env that sets up ruby (via rbenv) as well as podman.  This means you don't need any local dependencies installed other than nix/flox, and your environment is completely isolated to this repository.

I've tested this primarily on linux, ensuring that a clean flox env can come up and run `rbenv exec bundle exec cookstyle` as well as a kitchen test.

### DIFF
--- a/.flox/.gitattributes
+++ b/.flox/.gitattributes
@@ -1,0 +1,1 @@
+env/manifest.lock linguist-generated=true linguist-language=JSON
--- a/.flox/.gitignore
+++ b/.flox/.gitignore
@@ -1,0 +1,5 @@
+run/
+cache/
+lib/
+log/
+!env/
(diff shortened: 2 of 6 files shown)


apenney:
It doesn't need nixos, it just uses nix.  So if you can install nix itself in your distro, you can install flox on top.  On macos it comes bundled, so you can just `brew install flox` and off you go.  On cachyos, which I use, I did a pacman -S nix and then did:

```
nix profile install \
      --experimental-features "nix-command flakes" \
      --accept-flake-config \
      'github:flox/flox/latest'
```

They have a container (ghcr.io/flox/flox) and rpm/deb packages too, so it's relatively easy to install across almost everything.

PR #844, opened 2026-04-17
Update OSM tile URL

danieldegroot2:
Updated OSM tile URL in remaining files. See https://github.com/openstreetmap/operations/issues/737

### DIFF
--- a/cookbooks/web/files/default/static/openlayers/OpenStreetMap.js
+++ b/cookbooks/web/files/default/static/openlayers/OpenStreetMap.js
@@ -14,9 +14,7 @@ OpenLayers.Layer.OSM.Mapnik = OpenLayers.Class(OpenLayers.Layer.OSM, {
      */
     initialize: function(name, options) {
         var url = [
-            "https://a.tile.openstreetmap.org/${z}/${x}/${y}.png",
-            "https://b.tile.openstreetmap.org/${z}/${x}/${y}.png",
-            "https://c.tile.openstreetmap.org/${z}/${x}/${y}.png"
+            "https://tile.openstreetmap.org/${z}/${x}/${y}.png"
         ];
         options = OpenLayers.Util.extend({
             numZoomLevels: 20,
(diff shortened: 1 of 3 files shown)


cookbooks/web/files/default/static/openlayers/OpenStreetMap.js
```diff
@@ -14,9 +14,7 @@ OpenLayers.Layer.OSM.Mapnik = OpenLayers.Class(OpenLayers.Layer.OSM, {
      */
     initialize: function(name, options) {
         var url = [
-            "https://a.tile.openstreetmap.org/${z}/${x}/${y}.png",
-            "https://b.tile.openstreetmap.org/${z}/${x}/${y}.png",
-            "https://c.tile.openstreetmap.org/${z}/${x}/${y}.png"
+            "https://tile.openstreetmap.org/${z}/${x}/${y}.png"
```
tomhughes: Does this still need to be an array if there's only one value? I suspect OL takes a string or an array?

cookbooks/web/files/default/static/openlayers/OpenStreetMap.js
```diff
@@ -14,9 +14,7 @@ OpenLayers.Layer.OSM.Mapnik = OpenLayers.Class(OpenLayers.Layer.OSM, {
...
         ];
```
danieldegroot2:
```suggestion
        var url = 'https://tile.openstreetmap.org/${z}/${x}/${y}.png',
```

@tomhughes how does this look?
(ignore mixed diff, will make change manually)

tomhughes: Sure that's what I meant.

tomhughes:
Can you squash those last two commits together please, then we can get this merged.

danieldegroot2:
Superseded by #845.

tomhughes:
You could just have force pushed this PR to update it instead of opening a new new...

danieldegroot2:
As I mentioned in new PR, I only use the github.com website itself on my internet browser. I don't use any coding/console/app/etc.. (afaik the website does not have this option)
Regardless this comment should clarify what I mean by 'the browser version' in this case.

PR #846, opened 2026-04-21
Add second GPS database  for dev server

Rub21:
@Firefishy This PR adds a second GPS database to the dev server, as the  suggestion in the OPS meeting. Both databases run on the same PostgreSQL server and share the same user. The rails_port resource now accepts a gps_database_name, and the database.yml template was updated to use Rails multi-database config with primary and gps connections. The dev recipe creates and drops the apis_{name}_gps database for each site, with PostGIS and btree_gist extensions. It also adds a new rub21 site pointing to my gps_db branch to test the setup.

About the database.yml edit: this may need to be reset my changes, since it's allowed  that only the maintainer can edit it.

### DIFF
--- a/cookbooks/dev/recipes/default.rb
+++ b/cookbooks/dev/recipes/default.rb
@@ -360,6 +360,7 @@ if node[:postgresql][:clusters][rails_cluster.to_sym]
 
   node[:dev][:rails][:sites].each do |name, details|
     database_name = details[:database] || "apis_#{name}"
+    gps_database_name = "#{database_name}_gps"
     site_name = "#{name}.apis.dev.openstreetmap.org"
     site_directory = "/srv/#{name}.apis.dev.openstreetmap.org"
     log_directory = "#{site_directory}/logs"
@@ -386,6 +387,18 @@ if node[:postgresql][:clusters][rails_cluster.to_sym]
       postgresql_extension "#{database_name}_postgis" do
         cluster rails_cluster
         database database_name
+        extension "postgis"
+        owner "postgres"
+      end
+
+      postgresql_database gps_database_name do
+        cluster rails_cluster
+        owner "apis"
+      end
+
+      postgresql_extension "#{gps_database_name}_postgis" do
+        cluster rails_cluster
+        database gps_database_name
         extension "postgis"
         owner "postgres"
       end
@@ -435,6 +448,9 @@ if node[:postgresql][:clusters][rails_cluster.to_sym]
         database_port node[:postgresql][:clusters][rails_cluster.to_sym][:port]
         database_name database_name
         database_username "apis"
+        gps_database_port node[:postgresql][:clusters][rails_cluster.to_sym][:port]
+        gps_database_name gps_database_name
+        gps_database_username "apis"
         email_from "OpenStreetMap <web@noreply.openstreetmap.org>"
         gpx_dir gpx_directory
         log_path "#{log_directory}/rails.log"
@@ -571,6 +587,11 @@ if node[:postgresql][:clusters][rails_cluster.to_sym]
       end
 
       postgresql_database database_name do
+        action :drop
+        cluster rails_cluster
+      end
+
+      postgresql_database gps_database_name do
         action :drop
         cluster rails_cluster
       end
--- a/cookbooks/web/resources/rails_port.rb
+++ b/cookbooks/web/resources/rails_port.rb
@@ -41,6 +41,11 @@ property :database_port, String
 property :database_name, String
 property :database_username, String
 property :database_password, String
+property :gps_database_host, String
+property :gps_database_port, String
+property :gps_database_name, String
+property :gps_database_username, String
+property :gps_database_password, String
 property :email_from, String
 property :messages_domain, String
 property :gpx_dir, String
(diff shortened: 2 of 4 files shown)


tomhughes:
Sorry I was intending to look at adding the necessary chef support myself but I'll try and review this when I get a chance.

cookbooks/dev/recipes/default.rb
```diff
@@ -574,6 +596,11 @@
         action :drop
         cluster rails_cluster
       end
+
+      postgresql_database "#{database_name}_gps" do
```
pablobm, line 600: Can this use the existing variable?

```suggestion
      postgresql_database gps_database_name do
```

tomhughes: It can, yes.

Rub21: Done, thanks, I've  moved the variable `gps_database_name`  up so the variable is available in both the create and drop blocks.

cookbooks/dev/recipes/default.rb
```diff
@@ -376,6 +377,25 @@
...
+        extension "btree_gist"
+        owner "apis"
+      end
+
+      postgresql_extension "#{gps_database_name}_postgis" do
+        cluster rails_cluster
+        database gps_database_name
+        extension "postgis"
+        owner "postgres"
+      end
```
tomhughes: I'd put all this after the extensions are added to the main database, so that we group all the main database stuff together and then all the gps database stuff.

cookbooks/dev/recipes/default.rb
```diff
@@ -376,6 +377,25 @@
         owner "apis"
       end
 
+      postgresql_database gps_database_name do
+        cluster rails_cluster
+        owner "apis"
+      end
+
+      postgresql_extension "#{gps_database_name}_btree_gist" do
```
tomhughes: Is the gist extension actually needed for the gps database?

cookbooks/web/resources/rails_port.rb
```diff
@@ -171,7 +172,8 @@
               :port => new_resource.database_port,
               :name => new_resource.database_name,
               :username => new_resource.database_username,
-              :password => new_resource.database_password
+              :password => new_resource.database_password,
+              :gps_database_name => new_resource.gps_database_name
```
tomhughes: The `database` here is kind of redundant so I think we should drop it for consistence with the main database:

```suggestion
              :gps_name => new_resource.gps_database_name
```

I think we should also have a separate username and password for the gps database - even if they're the same on dev they may well not be in production.

cookbooks/web/templates/default/database.yml.erb
```diff
@@ -1,35 +1,77 @@
...
+<% if @port -%>
+    port: <%= @port %>
+<% end -%>
+    database: <%= @gps_database_name %>
+    username: <%= @username %>
+<% if @password -%>
+    password: <%= @password %>
+<% end -%>
+    encoding: utf8
+    connect_timeout: 3
```
tomhughes, line 35: Why do we need a connect timeout here? We don't have one for the main database?

tomhughes (changes requested):
In addition to the inline comments, the last three commits should be squashed together - one is just a fixup and the other is a change to the template to match the change made to it's usage in the previous commit.

Then I think the two remaining commits should be reversed, so that we add support for the second database and then a branch that uses it.

cookbooks/dev/recipes/default.rb
Rub21: Done, I've moved the gps database block after the main database extensions.

cookbooks/dev/recipes/default.rb
Rub21: You are right, it is not needed. I was just mirroring the extensions from the main database,  I've just removed.

cookbooks/web/resources/rails_port.rb
Rub21: Done, i have renamed to `:gps_name` and added separate username/password. Do you think we should also add host and port?

cookbooks/web/templates/default/database.yml.erb
Rub21: Since the gps database is secondary,  the  short timeout makes the gps connection fail fast so processes don't hang waiting , seems  the main database does not need it because if  it is down the site is down anyway

Basically it's to protect the main site,  if the gps database fails, only the gps section would be down, not the main functionality. So it would just trigger a warning instead of taking everything down.

image


I have document  the case here:  https://hackmd.io/pnjhFQYkTEmaGPJZiTlLNw#--What-happens-if-the-GPS-database-is-not-available

cookbooks/web/resources/rails_port.rb
tomhughes: We probably should actually - it's not essential right now but we'll likely want them for production.

cookbooks/dev/recipes/default.rb
```diff
@@ -390,6 +391,18 @@
         owner "postgres"
       end
 
+      postgresql_database gps_database_name do
+        cluster rails_cluster
+        owner "apis"
+      end
+
+      postgresql_extension "#{gps_database_name}_postgis" do
```
tomhughes, line 399: I realise I asked about btree before but not about this... It's not actually needed currently but I guess you are maybe considering using it for the points?

Rub21: Yes, we are considering it for the points/traces, would be good to have it ready in the DB. I can remove it and add it later in a separate PR if that is the right way.

cookbooks/web/resources/rails_port.rb
Rub21: Got it, I will add host and port for the gps database too.

Rub21: Done, I have added host and port for GPS database connection

tomhughes (approved):
I think this looks good now, it just needed the fixup commit squashing into the first one which I've just done.

tomhughes:
This is currently failing to deploy because your web site branch has issues with the gems it is trying to install - you need to do a bundle install to update the lock file I think.

Rub21:
I updated the gps_db branch: merged latest master and ran bundle install.  Gemfile.lock is now in sync. but seem it is necessary more work to do in the configs  to support two database  https://github.com/Rub21/openstreetmap-website/tree/gps_db

PR #849, opened 2026-05-07
Give source dir to prometheus metric script

joto:
For upcoming changes to the metric script that also needs data from the sources directory.


tomhughes:
Looks good to me, thanks.

PR #855, opened 2026-06-04
Add rub22 and rub23 branches to dev server

Rub21:
I'm adding these two dev sites to test the GPX compression and the new visibility levels.

- Compression of GPX files: https://github.com/openstreetmap/openstreetmap-website/pull/7124

- Visibility levels: https://github.com/openstreetmap/openstreetmap-website/issues/7089 and https://github.com/Rub21/openstreetmap-website/tree/simplify-gps-visibility

It would be good to deploy these branches,  so we can test the changes in a real scenario with large GPX data.

### DIFF
--- a/roles/dev.rb
+++ b/roles/dev.rb
@@ -134,11 +134,11 @@ default_attributes(
           :repository => "https://github.com/rub21/openstreetmap-website.git",
           :revision => "gps_db"
         },
-        :rub22 => {
+        :gpsvisibility => {
           :repository => "https://github.com/rub21/openstreetmap-website.git",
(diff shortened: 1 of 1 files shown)


tomhughes:
We should call these `gpsvisibility` and `gpxcompression` or something, not `rub22` and `rub23`...

Rub21:
I renamed, I thought that was aligned more with user names, but if we can use the functionality name, that sounds fine.

tomhughes:
Deployed, plus I renamed the existing one to `gpsdatabase`.

PR #856, opened 2026-06-18
db: Skip backup of gps_points table data

Firefishy:
Skip the backup of `gps_points` table data.

The `gps_points` table data could be re-imported from the source GPX files.

The `gps_points` table data bloats the backup size and directly delays the publishing of the planet data.

### DIFF
--- a/cookbooks/db/templates/default/backup-db.erb
+++ b/cookbooks/db/templates/default/backup-db.erb
@@ -5,6 +5,6 @@
 D=`date +%Y-%m-%d`
 F=/store/backup/osm-${D}.dmp
 
-pg_dump --user=backup --format=custom --file=$F openstreetmap && \
+pg_dump --user=backup --format=custom --exclude-table-data=gps_points --file=$F openstreetmap && \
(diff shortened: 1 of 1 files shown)


Firefishy:
For consideration if we want to do this...

pnorman:
I'm fine with this. pg_dump isn't for backups anyways, that's what a base backup + WAL is for.

Firefishy:
> I'm fine with this. pg_dump isn't for backups anyways, that's what a base backup + WAL is for.

We don't currently do base backups.

tomhughes:
If we move to base backups then we can go further than this PR and exclude all tables (like the history tables!) that the planet dump doesn't need.

pnorman:
> exclude all tables (like the history tables!)

Unfortunately it needs those for the history PBF :(

tomhughes:
Now that we have a successful base backup I have deployed an improved version of this that only dumps the tables that planetdump needs.

In particular this allows us to exclude all the current tables as they aren't used - the non-history dump actually uses the most recent version in the history tables rather than the current tables.

PR #860, opened 2026-07-19
Set referrer-policy on embed page

pnorman:
I wasn't able to properly test this since we block access from 127.0.0.1


tomhughes:
What's the background here? Why do we want to change it?

pnorman:
> What's the background here? Why do we want to change it?

We want to be able to get an idea of the normal osm.org traffic vs embed traffic on the tile CDN. I know it won't be perfect since some browsers will ignore it, but right now we have no idea of the scale of the two. It's been an action item on me for the past few ops meetings

